PR #181, opened 2014-01-17
rewrite rules should apply to .php paths too

artooro:
If a php web app path uses .php and expects those URLs to be rewritten it would not work.
This change applies the main rewrite rules to .php locations as well.

This fixes a problem when running concrete5.


d00p:
is this with the current version (0.9.33.1) still an issue?

PR #214, opened 2015-01-05
Update main.css

Commifreak:
Could someone review this? float: right; breaks the layout:

Before:

bildschirmfoto 2015-01-05 um 01 08 51

After:
bildschirmfoto 2015-01-05 um 01 08 59

Maybe some other adaptions are needed..?

### DIFF
--- a/templates/Sparkle/assets/css/main.css
+++ b/templates/Sparkle/assets/css/main.css
@@ -298,7 +298,6 @@ aside.right {
 }
 
 .overviewsearch {
-	float:right;
 	font-size:80%;
 	text-align:right;
 }


d00p:
Sounds like it's related to http://redmine.froxlor.org/issues/1466 - need to investigate

d00p:
Unable to reproduce with current-git source...are you sure this is still a problem? Or are you using 0.9.32-stable?

Commifreak:
Argh - not tested. Yep, I use 0.9.32-stable.

Just tested your fix: It resolves the issue. So, sorry for wasting your time ;)

PR #218, opened 2015-01-10
Update dutch.lng.php - Please check

Lednerb:
Moved & Commented doubled entries.
@scarya @adlerweb and all dutch friends, please check which versions are the better ones.


Lednerb:
The uncommented versions were the newer ones.

PR #222, opened 2015-01-26
Create usr_local_etc_pure-ftpd.conf

heaven73:
template for freebsd and pure ftp daemon - tested on freebsd 10.1 working


d00p:
Can this please be put together with #223 + add missing commands to install the package (like we do for the other daemons too)

PR #223, opened 2015-01-26
Create usr_local_etc_pureftpd-mysql.conf

heaven73:
tested and working on freebsd 10.1


d00p:
Can this please be put together with #222 + add missing commands to install the package (like we do for the other daemons too)

PR #224, opened 2015-01-26
Pure-FTPd template for FreeBSD

heaven73:
Tested on FreeBSD 10.1


heaven73:
Diesmal hoffentlich klappts!

PR #225, opened 2015-01-31
Bug #1479

RipClaw2971:
If you login with a false password this password will be set if the hash algorithm differs from the one used on the password in the database.

Solution: Only update the password hash in the database if the password matches


d00p:
Fixed already in 0.9.33-rc3 which will be online in a second

PR #233, opened 2015-03-23
VHost configs templates

hpmewes:
New feature to show, create, add and delete VHost config templates.
screenshot 1
screenshot 4
screenshot 5
screenshot 6

In my case it use it to create a nginx VHost config template for silverstripe installations. I need the same config for some domains. Before i added the feature the nginx .conf files must edited manually to add some location directives and so one. The recent problem was the unix socket which was not passed by froxlor now it's passed to template via {SOCKET}. 
The VHost config file can be selected in domain section.
screenshot 2

The second thing what i changed is that now the default directives 
location / { ... } 
location @rewrites { ... } 
location ~ ^(.+?\.php)(/.*)?$ { ... }
could be removed in vhost config by checking follow.
screenshot 3

I think that this feature was wished from many users from community and hope you will enjoy it and publish as soon is possible.
greetings from germany.


d00p:
Interesting..need to check. First impression is "why is there only a replacer for the SOCKET and not for CUSTOMER_DOCROOT / DOMAIN_DOCROOT etc.pp." but I need to read through the code first to know why this is not needed...

Thanks so far

d00p:
Also: is this really only for nginx? Pity

hpmewes:
in my case i use it only for nginx. i think it can be used in some other cases but should be modified

hpmewes:
with the other webserver configurations like lighthttpd apache i'm not similar.
i hope it's although usefull for you

chrisv2:
@hpmewes I believe PR #244 would also address your problem, could you test this?

hpmewes:
Hello,

thx for your comment. i will check it next time. i'll give you feedback then.

d00p:
Why not store $vhost_usedefaultlocation within the template and not the domain? It's most likely the template that requires these directives or not...not the domain itself...

d00p:
has been merged but does in no way work as expected

PR #234, opened 2015-04-04
opensuse 13.1 xml config template

wrosenauer:
Created an initial openSUSE 13.1 xml config template.
Why 13.1?
- because for now we had an 11 based one which is not supported anymore
- because 13.1 will be Evergreen supported and therefore has a slightly longer lifetime than
  current release 13.2
- 13.1 should be fairly similar to 13.2 anyway and also compared to SLES12


d00p:
Could you please provide a CLEAN pull-request just for the XML? We do not want other changes to interact here

PR #236, opened 2015-04-28
Add option to use webserver group for home and domains

chrisv2:
While adding the webserver to the user groups is a clever hack, this won't work if the customer dir is mounted over NFS. This patch adds a setting where one can opt to set the group owner to the web server instead, as originally suggested in #532.

The problem with files uploaded over ftp or generated by php having the wrong group is solved by setting the [setgid](http://en.wikipedia.org/wiki/Setuid#setuid_and_setgid_on_directories) bit for these folders.

The new setting has been added to the security section of the config, because I thought it would fit best there. Of course, if you prefer to have it somewhere else, I can change this respectively.


install/updates/froxlor/0.9/update_0.9.inc.php
```diff
@@ -2976,3 +2976,12 @@
         updateToVersion('0.9.34-dev3');
     }
 }
+
+if (isFroxlorVersion('0.9.34-dev3')) {
+
+    showUpdateStep("Updating from 0.9.34-dev3 to 0.9.34-dev4", false);
+
+    showUpdateStep("Adding new setting for customer dir group");
+    Database::query("INSERT INTO `panel_settings` SET `settinggroup` = 'system', `varname` = 'customerdir_group_webserver', `value` = '0';");
```
d00p: please use "Settings::addNew('group.varname', 'value');"

d00p:
Question is: why doesnt it work when the webserver-user is already in the customer-users group? Having the customer-homedirs mounted via NFS is not quite the "ordinary way"...pretty special. Also, you don't explain the setting. You write what it does, but don't describe why one would need this.

I don't really see the need right now; comments are welcome

chrisv2:
Short answer: It does not work because the NFS server does not know that the webserver is in all these groups.

Long answer: Originally, NFS did consider group membership data on the client system, but unfortunately this is limited to 16 groups. Since this is not very useful, most modern systems run their NFS servers with the [` --manage-gids`](http://www.xkyle.com/solving-the-nfs-16-group-limit-problem/) option, which means that the NFS server does not regard the group memberships sent by the client, but uses its local group data instead. This solves the 16-group-problem, but now the NFS server must know the group memberships, which is impractical in my scenario (see below).

Why would one use NFS in the first place? Because it is more flexible, faster (compared to a filesystem in a qcow2 or raw VM image), and allows to separate production/customer data from code (think Docker), therefore simplifying backups and maintenance. I agree that this is probably overkill if you only have a single server or VM with all customers on it, but once you go beyond that, NFS is pretty common. And then you don't want to add the libnss_mysql stuff to the NFS server, and without that, it wouldn't work, as described above.

Regarding the description, do you think its sufficient to add "Only check this if your customer directory is served over NFS" to the label text, or should I create a wiki page and link that?

Cellebyte:
Load balancing with multiple app servers for one domain
This means that two or more nodes actually have the same configuration, just with different IPs. We can either add two A-records for these IPs in the DNS, or we can put a balancer in front (configuring the balancer itself is out of the scope of this document). Of course, in this case we must make sure that the customer directory is the same for all nodes. This can be achieved either by using a distributed file system, or by mounting the customer directory over NFS (see pull request #236). 
Need it! NFS is not working!

BNoiZe:
@Cellebyte This patch is nearly two years old (April 2015) and wont be merged at the current state of development. If you really need this patch consider patching froxlor for yourself.

Cellebyte:
@BNoiZe I knew that. I must fix it for my own but this option is not unnecessary. Run into the same problem with service docker container which use the same NFS file server. For load balancing and software defined infrastructure it is a must have to scale by adopting more service container using the same data. Is their any other way to do so?

Slydder:
I've been clustering Froxlor for a few years now with great results. 
There is only 1 change needed to actually get froxlor working correctly 
in a clustered situation (listening on all IP addresses allows for 
stress free clustering) if behind a HA loadbalancer.

Send me a message if you would like to learn a bit more about it. You 
can also read up on a few things I am working on for Froxlor.

https://www.itadmins.net/2016/06/13/745/#more-745 - HA Clustering panel 
plugins
https://www.itadmins.net/2015/10/12/fix-for-ha-clustering-froxlor/#more-638 
- Froxlor HA Clustering
https://www.itadmins.net/2015/09/11/froxlor-love/#more-627 - Thoughts on 
Frox

And if you are wanting to NFS mount your content/configs then consider a 
move to GlusterFS. If you do then the following may be useful.

https://www.itadmins.net/2016/03/18/getting-useful-metrice-from-glusterfs-into-grafana-using-diamond/#more-678

Chuck

d00p:
I have a froxlor setup with split services on different servers and storage mounted via NFS - never needed this PR or any special hack, works fine. Close

PR #197, opened 2014-07-09
Add possibility to have external database server for customers database

ffrewer:
Fixes Bug #1438: Second External database server for customers database
not anymore possible

PR #237, opened 2015-04-29
allow external customer database, fixes #1438

chrisv2:
Patch based on PR #197, adapted to master, with some additions.
## Description

This patch allows to have a separate customer database (or even multiple customer databases). To enable it, just add a second database to `$sql_root` in `userdata.inc.php`:

```
$sql_root[0]['caption']='Default';
$sql_root[0]['host']='127.0.0.1';
$sql_root[0]['user']='root';
$sql_root[0]['password']='xyz';
$sql_root[0]['hidden']=true;

$sql_root[1]['caption']='Customer';
$sql_root[1]['host']='10.0.10.18';
$sql_root[1]['user']='root';
$sql_root[1]['password']='abc';
```

Now, when the customer wants to create a new database, he will be able to select the server from a dropdown. If you don't want to allow customer databases on the Froxlor db host itself, you can set the main database to `hidden` (see above) - then it will not be available in the customer panel.

The patch also enables using wildcards in `system.mysql_access_host`, which was not possible before (but is useful, if you want to limit access to a network or number of hosts).
#### Implementation notes
- the host for a given customer db is stored in `panel_databases`.`dbserver`, which is an index to `$sql_root`. This field was already there, and is now used again.
- to operate on a specific database, call `Database::needRoot()` with the server's index as the second parameter
- If set, `$linkWithoutDatabase` removes the database name from the PDO DSN, which is required if `$dbserver` is not `0` (this can probably be removed completely, but it was in the original patch, so I left it in)
- `correctMysqlUsers()` has been rewritten completely, as the old version was a total mess and would delete non-froxlor database users under some circumstances.


.gitignore
```diff
@@ -6,3 +6,5 @@ install/update.log
 .settings/
 *.diff
 *~
+.idea/
```
chrisv2: While not strictly belonging to this PR, this is a useful addition for PHPStorm users...

chrisv2:
@d00p did you see this PR? Any comments?

d00p:
I have seen it. Github sends notifications. No time yet.

d00p:
outdated

PR #239, opened 2015-05-11
Bug #1533

RipClaw2971:
"Require all granted" wurde nicht gesetzt wenn man mod_proxy_fcgi verwendet

### DIFF
--- a/scripts/jobs/cron_tasks.inc.http.15.apache_fcgid.php
+++ b/scripts/jobs/cron_tasks.inc.http.15.apache_fcgid.php
@@ -48,6 +48,16 @@ class apache_fcgid extends apache
 					$php_options_text.= '  SetHandler proxy:unix:' . $php->getInterface()->getSocketFile()  . '|fcgi://localhost'. "\n";
 					$php_options_text.= '  </FilesMatch>' . "\n";
 
+					$mypath_dir = new frxDirectory($domain['documentroot']);
+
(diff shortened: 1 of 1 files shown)


d00p:
Explain/show why this is necessary please. I dont have problems with it for months...

RipClaw2971:
The default configuration in /etc/apache2/apache2.conf in Debian Jessie is this:

<Directory />
        Options FollowSymLinks
        AllowOverride None
        Require all denied
</Directory>

There are two options to allow access to the websites. The first one is a global solution:

<Directory /var/customers/webs>
        Require all granted
</Directory>

The second is a "Require all granted" per virtual host.

d00p:
As said, i have this setup for months, no problem..don't see why this is necessary

RipClaw2971:
If you don't set a "Require all granted" for the directory the only thing you see is "Forbidden".
In the apache2.conf there are two other blocks to whitelist /var/www and /usr/share

```

<Directory /usr/share>
        AllowOverride None
        Require all granted
</Directory>

<Directory /var/www/>
        Options Indexes FollowSymLinks
        AllowOverride None
        Require all granted
</Directory>

```

I've got the problem yesterday on my new server and tracked the "Forbidden" down to the apache2.conf. The only working solution was to set a "Require all granted" to every virtual host to the document root.

d00p:
Well, additional "require"s are no harm - maybe it's just works for me on gentoo out of the box

PR #244, opened 2015-07-06
Variable substitution in vhost specialconfig

chrisv2:
## Description

This patch allows to use variables in domain-specific specialconfig. The following replacements are currently implemented:

Variable | Replacement
-----------|-----------------
`{DOMAIN}` | domain
`{CUSTOMER}` | login name (=system user)
`{IP}` | ip
`{PORT}` | port
`{SCHEME}` | `https` if SSL is enabled in this vhost, `http` if not
`{DOCROOT}` | document root

## Use Case

You can use this for example to enable Python/WSGI globally, or per vhost:

```apache
WSGIDaemonProcess {DOMAIN}-{SCHEME} processes=2 threads=15 user={CUSTOMER} group={CUSTOMER}
WSGIProcessGroup {DOMAIN}-{SCHEME}

<Directory "{DOCROOT}">
    ...
    <FilesMatch "\.wsgi$">
      SetHandler wsgi-script
      Options +ExecCGI
    </FilesMatch>
    ...
</Directory>
```


scripts/classes/class.HttpConfigBase.php
```diff
@@ -0,0 +1,38 @@
...
+	 * @param $template
+	 * @return string
+	 */
+	protected function processSpecialConfigTemplate($template, $ipandport, $domain) {
+		$templateVars = array(
+			'DOMAIN' => $domain['domain'],
+			'CUSTOMER' => $domain['loginname'],
+			'IP' => $ipandport['ip'],
+			'PORT' => $ipandport['port'],
+			'IS_SSL' => $domain['ssl'],
```
d00p: $domain['ssl'] won't be an empty string, it's either 1 or 0, also, whether it's a ssl-vhost should be read from whether the currently used ip/port is SSL=Yes.

And i don't get what a '1' for enabled SSL is useful for except for your very special case to distinguish between ssl and non-ssl as you cannot use IF's there

d00p:
Basically a really good job for the needs but I'm missing the info on the frontend for the users

chrisv2:
Added some description. If you want me to, I can put longer description into the Wiki, then we could link that.

d00p:
I'm still unsure about the behavior of "IS_SSL"...why is its value 'ssl' when activated and not for example 'true' or '1'? How is this usable without the possibility of an IF-statement? Some directives may require the value to be 'ssl' others might need completely different stuff...

chrisv2:
I admit this is a little bit hacky, but actually being able to differentiate between https and http is the main purpose of the patch ;)

Eventually it would make more sense to have a variable `{SCHEME}` (or `{PROTOCOL}`)  instead which always evaluates to either `http` or `https`?

d00p:
{SCHEME} seems to be helpful in many situations;

chrisv2:
replaced `{IS_SSL}` with `{SCHEME}` and merged from master

PR #245, opened 2015-07-28
Add foreign key constrains to database

chrisv2:
This PR adds foreign key constraints to the database, and additionally normalizes the values for `panel_domains.parentdomainid` and `mail_virtual.popaccountid` (both columns do allow and use `NULL` now for empty values, instead of `0`, which would violate the foreign key constraint).

#### Notes
* all tables are switched from the (deprecated) MyISAM engine to InnoDB, which is the recommended engine nowadays (MyISAM does not recognize foreign key constraints)
* some foreign key fields have been switched from `int(11)` to `int(11) unsigned`, because for foreign key constraints, the type must match the key of the referred table
* while this patch does not add new functionality in itself, it is a necessary groundwork for adding an ORM later (which would allow us to get rid of a lot of cruft in the long run)
* Another nice feat of referential integrity is that you can move IDs now easily - this is useful if you want to join two or more Froxlor databases, for example. Just update, for example, the `customerid` for some customer, and `ON UPDATE CASCADE` will update all the related fields as well.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,5 +4,6 @@ install/update.log
 .buildpath
 .project
 .settings/
+.idea
 *.diff
 *~
--- a/admin_customers.php
+++ b/admin_customers.php
@@ -77,7 +77,7 @@ if ($page == 'customers'
 					SELECT COUNT(`id`) AS `domains`
 					FROM `" . TABLE_PANEL_DOMAINS . "`
 					WHERE `customerid` = :cid
-					AND `parentdomainid` = '0'
+					AND `parentdomainid` IS NULL
 					AND `id`<> :stdd"
 				);
 				Database::pexecute($domains_stmt, array('cid' => $row['customerid'], 'stdd' => $row['standardsubdomain']));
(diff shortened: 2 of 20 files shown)


d00p:
Very nice patch, i hope it's been tested? :)

chrisv2:
Am 04.08.2015 um 14:03 schrieb Michael Kaufmann:
> Very nice patch, i hope it's been tested? :)

well, the unittests didn't fail so I figure it's okay ;)

just kidding, I did some testing and I do use the migrated DB in my dev
system right now, so far without problems. I did add/edit/delete some
domains, aliases and mail accounts, only got some minor glitches which
were fixed in the subsequent commits.

chrisv2:
66e169c adds some important improvements:
* `Database::query()` seems to silently ignore errors when multiple SQL statements are contained in one function call. Now we only execute a single statement per call, so we get errors when something fails
* remove rows pointing to deleted customers/domains in `panel_traffic`, `panel_diskspace` and `domain_redirect_codes` - these were found in some of our Froxlor installations, for whatever reason
* allow `NULL` on `panel_domains.adminid` (for subdomains created by customers)
* allow `NULL` on `panel_domains.ismainbutsubto` and add a foreign key (for subdomains created by admins, see issue 329)
* allow `NULL` on `panel_tickets.customerid` (this is true for replies to customer tickets)
* force user to explicitely confirm that a database backup exists before upgrading, in case something goes wrong

PR #247, opened 2015-08-16
Fixed composeSslSettings 

Tak0r:
Since the composing of the ssl settings got seperated it still used variables not available in the function scope and broke the regeneration of the ip/ports config files.


d00p:
Ah the file-exist checks where copies from somewhere else, hence the different variable-names, thx for finding this :)

Tak0r:
NP, keep up the good work.

PR #250, opened 2015-09-09
Fixing some (big) peanuts

blueJack92:
Fixing permission problems with files, which have 0640 as file attributes.
Fixing problems in main.cf (Postfix). setgid_group must be set, otherwise running newaliases will fail.
Moreover I set the missing paths correctly in the main.cf (regard to Debian Jessie).


d00p:
Why only fixing jessie? What about the other distros?

blueJack92:
I only tested these with Jessie. Maybe other distros have no problems - don't know. :-)

PR #254, opened 2015-10-29
Updated CipherSuites...

Commifreak:
... due to incompatibility between Server <-> Java

With the current set of Suites, Java returns: javax.net.ssl.SSLHandshakeException: Received fatal alert: handshake_failure

I've updated the Suite as suggested by https://raymii.org/s/tutorials/Strong_SSL_Security_On_Apache2.html (scroll down to "The Cipher Suite").

This results to: https://www.ssllabs.com/ssltest/analyze.html?d=git.kluthr.de and enables Java-Clients (NetBeans in my case ;) ) to connect again.

The old suite seems to be stronger but makes Java-Clients unable to connect. Maybe someone have better solutions for this.


Commifreak:
This CipherSuite has even more compatibility (with Banking4 <-> ownCloud as example):

`EECDH+ECDSA+AESGCM:EECDH+aRSA+AESGCM:EECDH+ECDSA:EECDH:EDH+AESGCM:EDH:+3DES:ECDH+AESGCM:ECDH+AES:ECDH:AES:HIGH:MEDIUM:!RC4:!CAMELLIA:!SEED:!aNULL:!MD5:!eNULL:!LOW:!EXP:!DSS:!PSK:!SRP`

HolySephi:
I would not suggest to change the default cipher list. This is why the cipher list is configurable in the webinterface - if you need lesser strength you are free to set it up according to your needs.

The default should provide security in this case.

Commifreak:
Hm, sounds reasonable ;)

d00p:
@HolySephi is right

PR #263, opened 2015-12-02
fix path_info with try_files

Stricted:
and yes, its tested


d00p:
is that acutally still an issue?

Stricted:
i dont use froxlor anymore
so i have no idea

d00p:
so why answer to this then? lol....

PR #264, opened 2015-12-05
Add option to allow customers to specify filepaths for ssl instead of copying file contents

cobrafast:
Implemented my idea for very rudimentary Let's Encrypt support; letting customers specify file paths to ssl files directly rather than having them paste contents and froxlor create files. The new behaviour can be toggled with a checkbox in the global SSL settings.

I have tested this code and it seems to work as far as I can tell.

Additionally to the code in this pull request, a record needs to be inserted to the database:
```
INSERT INTO `panel_settings` (`settingid`, `settinggroup`, `varname`, `value`) VALUES (NULL, 'system', 'ssl_customers_set_paths', '0');
```

See https://redmine.froxlor.org/issues/1470


HolySephi:
But enabling this new setting forces each user to use paths am i correct?
If you could rework this feature to support both simultaniously this would be the perfect addition to the current ssl feature

cobrafast:
I'm not sure if automatic distrinction between a path and file contents is easily possible, since not only the ssl files, but the entire directory may be read protected, so `is_file()` might fail although the file exists.

Or should it just assume it's a path when the `openssl_x509_*()` tests fail?

HolySephi:
You could just check if the content starts with "-----" as in every certificate or the first (and only line) has dashes.
Maybe it's a bit more safe if you check if it's a path first and then fallback to cert mode like checking if its only one line and the line has at least one dash in it or something like that

You don't have to validate the cert or the file existance to detect the setting but be aware - if the file is missing your webserver won't start so your USER can break your whole setup.

d00p:
why bother customer with paths?! If that tool is somehow usable at all, we could just let the customer enable "create ssl cert for domain" via the panel and be done with it...

HolySephi:
but then you would need the full LE logic implemented in froxlor...

cobrafast:
>You could just check if the content starts with "-----" as in every certificate or the first (and only line) has dashes.

Checking if input starts with `/` or `-` may be a good solution. What do you think?

>You don't have to validate the cert or the file existance to detect the setting but be aware - if the file is missing your webserver won't start so your USER can break your whole setup.

That's not entirely true; Apache 2.4 will start with incorrect paths to SSL files, SSL just wont work on the affected VHosts.

d00p:
@HolySephi thought this was the point...customer's do not have shell access to create certs themselves

HolySephi:
This pull request is do get an easier integration of external certificates as an addition without the whole hassle of rewriting too much code

I like this idea because you are not limited to something like LE but you are free to specify the files for easier automation.

You would be able to integrate direct LE logic additionally if you want to

cobrafast:
@d00p I think that depends on how you use Froxlor. An admin could create the certificate for the customer and tell them the path; or like I do, some of the customers have shell access and Froxlor acts more as a tool for simplification to do frequent tasks. Thats why I implemented this feature as an admin-activated option, so people that don't want to give customers this power can simply opt to not enable it.

d00p:
a) i would not want to be available for customers 24/7 to create them certificates whenever they want
b) i won't give customers shell access on a shared-hosting environment, you are free to do that, but that's not our way

fritz-net:
I really like the idea behind the patch, but d00p is totally right. However some people like me only use froxlor for themselves to simplify there configuration where this is a good solution. Maybe it is an idea to implement this to be activate able per domain/customer.

I would not implement the whole letsencrypt logic in froxlor because you can simply query the cert by calling `./letsencrypt-auto certonly -a webroot --webroot-path /var/www/root_dir_of_subdomain/ -d sub.domain.tld` which is not very complicated to do with php I think

d00p:
There won't be an option to specify paths for customers as they can copy'n'paste (any) certificates content in the textareas and it will work. Anything else related to LetsEncrypt would be done using these  already existing fields.

PR #270, opened 2016-01-17
APCu Info page

hypernics:
Added an APCu info page to Server menu with clear cache function.
Based on https://github.com/krakjoe/apcu/blob/master/apc.php
Vertical image bars changed to horizontal.

APCu info page

### DIFF
--- a/admin_apcuinfo.php
+++ b/admin_apcuinfo.php
@@ -1,0 +1,407 @@
+<?php
+
+/*
+  +----------------------------------------------------------------------+
+  | APC                                                                  |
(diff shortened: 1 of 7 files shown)


d00p:
I don't quite see the point for including this...when running as fcgid or php-fpm only infos about the froxlor panel usage can be displayed, not of all customers/domains...is that really that important?

hypernics:
apcu is system wide (i use php-fpm). it always display all customers/domains usage.

d00p:
very well :)

PR #276, opened 2016-01-28
fix rand() invocation

hph86:
This fixes invocation of rand() which either takes 0 or 2 arguments.


d00p:
Using no arguments will generate a possible small value not long enough, better use 

``
rand(time(), getrandmax())
``

what do you think?

d00p:
fixed as of https://github.com/Froxlor/Froxlor/commit/2bb8f707ebfc19ceefb318617558e1faf102a5b8

PR #280, opened 2016-02-03
Added basic let's encrypt - supprt, fixes #1470

EleRas:


EleRas:
Valid question, I changed it

PR #284, opened 2016-02-09
add function getFormFieldDataText()

nachtgeist:
Previously webserver configs would contain CRLFs from
system.default_vhostconf on admin_settings.php.

This patch adds a new function which automatically gets recognized by
getFormFieldData() and mangles textarea form elements through
str_replace("\r\n", "\n", ...).

### DIFF
--- a/lib/functions/formfields/text/function.getFormFieldDataText.php
+++ b/lib/functions/formfields/text/function.getFormFieldDataText.php
@@ -1,0 +1,26 @@
+<?php
+
+/**
+ * This file is part of the Froxlor project.
+ * Copyright (c) 2010 the Froxlor Team (see authors).
+ *
+ * For the full copyright and license information, please view the COPYING
+ * file that was distributed with this source code. You can also view the
+ * COPYING file online at http://files.froxlor.org/misc/COPYING.txt
+ *
+ * @copyright  (c) the authors
+ * @author     Daniel Reichelt <hacking@nachtgeist.net> (2016-)
+ * @license    GPLv2 http://files.froxlor.org/misc/COPYING.txt
+ * @package    Functions
+ *
+ */
+
+function getFormFieldDataText($fieldname, $fielddata, &$input) {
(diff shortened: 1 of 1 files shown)


lib/functions/formfields/int/function.getFormFieldDataText.php
```diff
@@ -0,0 +1,28 @@
+<?php
+
+/**
+ * This file is part of the Froxlor project.
+ * Copyright (c) 2003-2009 the SysCP Team (see authors).
```
d00p: Not needed as this file was never part of SysCP

lib/functions/formfields/int/function.getFormFieldDataText.php
```diff
@@ -0,0 +1,28 @@
...
+ * Copyright (c) 2010 the Froxlor Team (see authors).
+ *
+ * For the full copyright and license information, please view the COPYING
+ * file that was distributed with this source code. You can also view the
+ * COPYING file online at http://files.froxlor.org/misc/COPYING.txt
+ *
+ * @copyright  (c) the authors
+ * @author     Florian Lippert <flo@syscp.org> (2003-2009)
```
d00p: Not needed as this file was never part of SysCP

d00p:
File is in the wrong folder, **formfields/int/** should be **formfields/text/** i guess?

nachtgeist:
Thanks, fixed.

(I noticed the formfields/text directory only after your comment...)

d00p:
Has his been tested so no other text-settings break?

nachtgeist:
I was surprised at first, too, not to find any other usages, but the only applicable call stack is

function.processForm.php:processForm(|Ex)() →
function.getFormFieldData.php:getFormFieldData() →
formfields/*/function.getFormFieldData*()

- processFormExt() only ever gets called from admin_settings.php
- processForm() is unused.

And since system_default_vhostconf is the only textarea in all of admin_settings: yes it's (completely) tested, no breakage.

PR #285, opened 2016-02-09
Polish Fix 

tczaude:
Small error fixed and re add polish translation somewhere gone in main repository


d00p:
missing language file for panel, this is just for the installation

PR #290, opened 2016-02-15
Rename Portuguese language file

stweil:
This fixes a typo in the name of the file (found by codespell).

Signed-off-by: Stefan Weil sw@weilnetz.de


EleRas:
I think this will cause a problem with existing installations, since then the file would be simply missing -> needs to be handled in updater as well

stweil:
Ah yes, that's correct. I now looked at `install/updates/froxlor/0.9/update_0.9.inc.php` (I assume that's the right place for the update). Should I add the database update to the last `if` block?

Before doing that, an unrelated bug in the updater should be fixed, see pull request #293.

d00p:
That's nothing too bad, cosmetics

PR #291, opened 2016-02-16
Logging

nachtgeist:


lib/classes/logger/class.FroxlorLogger.php
```diff
@@ -104,31 +104,8 @@ public function logAction ($action = USR_ACTION, $type = LOG_NOTICE, $text = nul
...
-					break;
-				case LOG_DEBUG:
-					$_type = 'debug';
-					break;
-				default:
-					$_type = 'unknown';
-					break;
-			}
-			echo "[".$_type."] ".$text.PHP_EOL;
+		if (self::$crondebug_flag || $type <= LOG_WARNING) { # more severe levels have smaller numbers
```
d00p: Please use // for comments, not #

d00p:
Has this been tested?

nachtgeist:
Worst case: small syntax errors in the nginx/lightning cronjobs during the "fwrite" changes, since I'm only using apache.

The rest worked fine though.

lib/classes/logger/class.FroxlorLogger.php
nachtgeist: updated.

lib/classes/logger/class.FroxlorLogger.php
```diff
@@ -104,31 +104,8 @@ public function logAction ($action = USR_ACTION, $type = LOG_NOTICE, $text = nul
...
-					break;
-				case LOG_DEBUG:
-					$_type = 'debug';
-					break;
-				default:
-					$_type = 'unknown';
-					break;
-			}
-			echo "[".$_type."] ".$text.PHP_EOL;
+		if (self::$crondebug_flag || $type <= LOG_WARNING) { // more severe levels have smaller numbers
```
d00p: the "|| $type <= LOG_WARNING" is wrong here since we're using the logger everywhere in the panel and therefore on every warning, error, etc. it will echo the message. I'll remove that in the merge

nachtgeist: Good catch. But why use `$type > 4` instead of `$type > LOG_WARNING`?

d00p: Just because i had that in there before your PR ;)

PR #293, opened 2016-02-16
Fix update process from Froxlor version 0.9.28-svn1

stweil:
A typo in the name of a local variable prevented database updates.
Fix also an unrelated whitespace issue in an empty code line.

Signed-off-by: Stefan Weil <sw@weilnetz.de>


stweil:
@d00p, all Froxlor installations which have missed that update (because of the typo) will never get it.
Should that database update also be added for the latest version, so everybody gets it?

d00p:
Can be done sure, gladly its nothing super important ;)

PR #295, opened 2016-02-17
Added Authcode to panel_domains

megaspatz:
- added authcode to table panel_domains
- added authcode to domain-add and -edit page for admins
- added authcode to domain-edit page for customers

### DIFF
--- a/install/froxlor.sql
+++ b/install/froxlor.sql
@@ -555,7 +555,7 @@ INSERT INTO `panel_settings` (`settinggroup`, `varname`, `value`) VALUES
 	('panel', 'password_numeric', '0'),
 	('panel', 'password_special_char_required', '0'),
 	('panel', 'password_special_char', '!?<>§$%+#=@'),
-	('panel', 'version', '0.9.35-dev5');
+	('panel', 'version', '0.9.35-dev6');
 
 
 DROP TABLE IF EXISTS `panel_tasks`;
--- a/install/froxlor.sql
+++ b/install/froxlor.sql
@@ -246,6 +246,7 @@ CREATE TABLE `panel_domains` (
   `add_date` int( 11 ) NOT NULL default '0',
   `registration_date` date NOT NULL,
   `termination_date` date NOT NULL,
+  `authcode` varchar(255) NOT NULL DEFAULT '',
   `phpsettingid` INT( 11 ) UNSIGNED NOT NULL DEFAULT '1',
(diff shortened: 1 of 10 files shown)


lib/formfields/customer/domains/formfield.domains_edit.php
```diff
@@ -101,7 +101,12 @@
 						'label' => $lng['domain']['openbasedirpath'],
 						'type' => 'select',
 						'select_var' => $openbasedir
-					)
+					),
+                                        'authcode' => array(
```
d00p: for the customer, i would hide this, if there is no authcode, e.g.
`visible => (!empty($result['authcode']))`

install/froxlor.sql
```diff
@@ -246,6 +246,7 @@ CREATE TABLE `panel_domains` (
   `add_date` int( 11 ) NOT NULL default '0',
   `registration_date` date NOT NULL,
   `termination_date` date NOT NULL,
+  `authcode` varchar(255) NOT NULL DEFAULT '',
```
d00p: this file also holds the current version number which needs to be adjusted (see values for panel_settings)

megaspatz: Ups, i forgot this. -> Done.

lib/formfields/customer/domains/formfield.domains_edit.php
megaspatz: Good idea -> done.

Guite:
If the field is hidden for customers, what is the value of this feature at all? Admins can also use their domain management provider to lookup the authcodes.

d00p:
@Guite who says it's hidden for customers? Also it's not even merged into our repository...

Guite:
@d00p okay, misunderstood the conversation above. The field is only hidden if there is no authcode available.

chaos234:
Would like to have this too in the next version or is there smth. wrong with it?

d00p:
Cant remember. I'll Check later.

d00p:
I'm not sure whether this makes sense as an Authcode is often only temporarily valid and there is no interface to any domain-registrar-service

chaos234:
It would make sense because tehn you can request the auth-code from your domain provider - e. g. hetzner - and past it in the interface for the customer and then the interface sends an automated message about an available auth code.

d00p:
I dont see any mailing code in this PR...

d00p:
three year old PR, sorry, close as no response

PR #299, opened 2016-02-20
Changes in the config templates

ianklemm:
Hello, 

I updated 2 things in the jessie config template:

- enabled plaintext auth because with no TLS enabled, you can´t login to the maildeamon
- disabled RLimitChroot because it brings some problems with it in shared hosting


d00p:
Please remove the rar file ;)

ianklemm:
Done

d00p:
And why do the changes only for Jessie?

ianklemm:
I will change it for the other distributions in the next minutes..

ianklemm:
In Wheezy is the mail config correct. It also uses an older version of proFTPd, so you don´t have to disable RLimitChroot.
In other distributions, I have fixed the error, if it was necessary.

PR #303, opened 2016-02-20
Nginx: Fix for redundantly inserted auth blocks

Churro:
The problem occurs if a Vhost is assigned multiple different auth names
in Froxlor. Each block is then added repeatedly, leading to an
unparseable configuration


d00p:
Are the multiple different auth names also a problem with apache / lighttpd?

Churro:
No, with Apache and lighttpd this does not happen since the only the last authname is taken there instead of repeatedly inserting the blocks.

PR #315, opened 2016-02-21
Nginx: Set default curve to secp384r1

Churro:
Why change the curve for ECDH key exchange from secp256r1 to secp384r1?
Because in the security world, we are practically heading towards RSA 4096-bit keypairs for TLS usage. An elliptic curve with primes of 256 bits relates to the cryptographic strength of a 3072-bits RSA key. Hence, for a sufficiently strong key exchange, we may also choose a curve for ECDH key exchange that is at least 384-bits. For a comparison, this corresponds to the equivalent of a 7680-bits RSA key or a 192-bits of symmetric key.

Basically, all browsers and servers supporting secp256r1 also provide secp384r1. Hence, compatibility issues are almost not existing. The change towards secp384r1 may involve only one downside: the used openssl implementation might not be as optimized as the one for secp256r1.


Churro:
Btw: By design, both curves are based on the Weierstrass equation. Nevertheless, secp384r1 is predominantly recommended by the Bettercrypto manual (https://bettercrypto.org/) and, opposed to secp256r1, gives 100% on the Qualsys SSLabs key exchange evaluation whereas secp256r1 scores only 90%.

PR #316, opened 2016-02-21
Feature: Test check the webserver config on reload

Churro:
**Problem:**
After adding a custom Vhost configuration, the webserver is reloaded. However, if the custom config contains a syntax error (missing bracket), no apparent error message is delivered.

**Solution:**
Having assembled the Vhost configuration, make use of the server's syntax checking features and emit an error message to the froxlor log if an error has been encountered.

Nginx and lighttpd tend to stop validation on the first error (one output line), Apache seems to also multiple syntax errors at once. Hence, in order to prevent stuffing a bunch of error messages into a single line, multiple log lines seem appropriate. 

This pull request also solves https://redmine.froxlor.org/issues/1434

### DIFF
--- a/scripts/jobs/cron_tasks.inc.http.10.apache.php
+++ b/scripts/jobs/cron_tasks.inc.http.10.apache.php
@@ -61,12 +61,7 @@ class apache extends HttpConfigBase {
 			if ($return == 127) {
 				continue;
 			} else if ($return > 0 && is_array($out) && count($out) > 0) {
-				// Always log this error, regardless of the settings
-				$log = Settings::Get('logger.log_cron');
-				Settings::Set('logger.log_cron', 1);
-
 				$this->logger->logAction(CRON_ACTION, LOG_ERR, 'configuration error: ' . implode("\n", $out));
-				Settings::Set('logger.log_cron', $log);
 			}
 
 			break;
@@ -278,7 +273,7 @@ class apache extends HttpConfigBase {
 					if ($row_ipsandports['ssl']) {
 						$srvName = substr(md5($ipport),0,4).'.ssl-fpm.external';
 					}
-					
+
 					// mod_proxy stuff for apache-2.4
 					if (Settings::Get('system.apache24') == '1'
 							&& Settings::Get('phpfpm.use_mod_proxy') == '1'
@@ -286,7 +281,7 @@ class apache extends HttpConfigBase {
 						$this->virtualhosts_data[$vhosts_filename] .= '  <FilesMatch \.php$>'. "\n";
 						$this->virtualhosts_data[$vhosts_filename] .= '  SetHandler proxy:unix:' . $php->getInterface()->getSocketFile()  . '|fcgi://localhost'. "\n";
 						$this->virtualhosts_data[$vhosts_filename] .= '  </FilesMatch>' . "\n";
-					
+
 					} else {
 						$this->virtualhosts_data[$vhosts_filename] .= '  FastCgiExternalServer ' . $php->getInterface()->getAliasConfigDir() . $srvName .' -socket ' . $php->getInterface()->getSocketFile() . ' -idle-timeout ' . Settings::Get('phpfpm.idle_timeout') . "\n";
 						$this->virtualhosts_data[$vhosts_filename] .= '  <Directory "' . $mypath . '">' . "\n";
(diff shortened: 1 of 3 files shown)


d00p:
I'd suggest making the syntax-check command a setting, like we do it with _webserver reload_, _cron reload_, _php-fpm reload_ etc.etc.etc.

d00p:
Also, config-testing **after** the reload command does not make much sense to me...

scripts/jobs/cron_tasks.inc.http.10.apache.php
```diff
@@ -52,6 +52,25 @@ public function reload() {
...
+		// Config test
+		$return = -999;
+		$cmds = [ 'apache2ctl', 'apache2', 'httpd2', 'httpd', 'apache' ];
+		foreach ($cmds as $cmd) {
+			$out = safe_exec($cmd . ' -t 2>&1 ', $return, [ '&','>' ]);
+			if ($return == 127) {
+				continue;
+			} else if ($return > 0 && is_array($out) && count($out) > 0) {
+				// Always log this error, regardless of the settings
+				$log = Settings::Get('logger.log_cron');
```
d00p: remembering the value of log_cron and setting it to 1 to log the action should not be necessary as warnings, errors, and critical messages **will** be logged anyway, see https://github.com/Froxlor/Froxlor/blob/master/lib/classes/logger/class.FroxlorLogger.php#L114

Churro:
Thanks for the suggestions.

Indeed, remembering the log level to force logging is no longer necessary since nachgeist's recent commit (https://github.com/Froxlor/Froxlor/commit/3d2b672feb2dcae6f93d8c21c44c4fc9cabde6fb). 
I've now added another commit, removing these statements.

- Concerning the **reload + check**:
Basically, the idea was to reload anyway and then check whether the reload might have been successful. Of course, it would be possible to first check and in case of success, reload. However, I didn't want the syntax check to be a possible Single Point of Failure because assuming that something weird happens there, e.g. the syntax test command doesn't work for whatever reason, then the webserver would never be reloaded. Aside from that, it makes no difference in practice whether the syntax is checked before or after the reload.

- Concerning the **syntax check command**:
I've thought about this also. However, for nginx and lighttpd there is always only this single command which should work in any situation. Hence, for these two you definitely need no separate command setting.

For Apache, this would be a legitimate option. Nevertheless, you then would have to set the exact command, valid for your Apache installation. In order to find out which command would work, you as an admin, have to try them manually within your shell. This is tedious, makes no sense and just adds a significant administration overhead. Not to forget that aside from the command, there is nothing to configure anyway.

Instead, by using the fixed preset of commands, the syntax check works out-of-the-box and can be introduced automatically using an ordinary Froxlor update.

d00p:
> Aside from that, it makes no difference in practice whether the syntax is checked before or after the reload.
Sure it does, the webserver is still reachable instead of down...that is a huge difference for me and my customers...

It doesn't matter if there is only one way the webserver checks...but what if people want to put a custom bash-script there or whatever doing other checks too or whatever. Why hardcode? Not necessary. Also what if i custom compiled my webserver and - for what ever reason - it's not within the PATH variable and i need to specify the whole path to the command?

> Instead, by using the fixed preset of commands, the syntax check works out-of-the-box and can be introduced automatically using an ordinary Froxlor update.

Well, our way is to add a new setting and ask the admin about its value when updating (preconfig + update). Sure it's nice to have an out-of-the-box working thing but I do not like to hardcore such commands

Churro:
I think we have to clarify a misunderstanding here:
If the webserver is reloaded with a faulty configuration file, it **does not **shut down. Instead, it simply continues with the previously working configuration. Nevertheless, you are not indicated whatsoever that the custom VHost configuration was not applicable.

Honestly, I do not see the point of making the path configurable. If you would enable the admin to specify a custom bash script, you would provide an attacker with admin access with a nice way to execute shell commands deliberately. The according output would even be logged to the froxlor log. 

A custom-compiled webserver which is not added to the PATH variable is probably an individual corner case, rather than a standard deployment. As of that, the **advantages** in terms of comfortability and less configuration overhead IMHO definitely outweighs a configuration option which has to be **specified manually every time**.

d00p:
Whatever, it is the froxlor way to let the admin specify commands for actions like that. Hardcoded commands are never a good solution.

d00p:
still hardcoded commands, and check after reload which makes absolutely no sense

PR #317, opened 2016-02-21
Feature: Enable multiple standard IPs

Churro:
**Problem:**
Until now, it was only possible to specify one IP address used for new domains by default. This is annoying if you run a dual-stack setup with IPv4 and IPv6 (cf. https://redmine.froxlor.org/issues/1509). 

**Solution:**
Make more than one IP address eligible by default. The solution has been tested working on many different hosts and is also suited to be applied as an update on-top of the existing setting.

### DIFF
--- a/admin_customers.php
+++ b/admin_customers.php
@@ -914,10 +914,10 @@ if ($page == 'customers'
 
 						// set ip <-> domain connection
 						$defaultips = explode(',', Settings::Get('system.defaultip'));
-						foreach ($defaultips as $defaultip) {
-							$ins_stmt = Database::prepare("
+						$ins_stmt = Database::prepare("
 							  INSERT INTO `" . TABLE_DOMAINTOIP . "` SET `id_domain` = :domainid, `id_ipandports` = :ipid"
-							);
+						);
+						foreach ($defaultips as $defaultip) {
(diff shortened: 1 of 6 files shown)


admin_customers.php
```diff
@@ -913,10 +913,13 @@
 						$domainid = Database::lastInsertId();
 
 						// set ip <-> domain connection
-						$ins_stmt = Database::prepare("
-							INSERT INTO `".TABLE_DOMAINTOIP."` SET `id_domain` = :domainid, `id_ipandports` = :ipid"
-						);
-						Database::pexecute($ins_stmt, array('domainid' => $domainid, 'ipid' => Settings::Get('system.defaultip')));
+						$defaultips = explode(',', Settings::Get('system.defaultip'));
+						foreach ($defaultips as $defaultip) {
+							$ins_stmt = Database::prepare("
```
d00p: The statement does not need to be prepared multiple times, once before the foreach would be enough

admin_customers.php
```diff
@@ -1272,10 +1275,13 @@
...
```
d00p: same here, it's enough to prepare the statement once and execute it multiple times

Churro:
Alright, optimized the prepared statement declarations in https://github.com/Churro/Froxlor/commit/ae2bad5ab4ca7f809604534e757340c0408b1dc2

PR #318, opened 2016-02-21
Feature: vHost templates (version 2)

Churro:
@hpmewes presented a nice implementation of pre-definable vHost templates in PR #233 .
Presumably due to some unresolved issues, his implementation has not been merged finally and remained as a pending pull request.

In this PR, a reworked implementation is provided with:
- bugfixes all over
- added support for apache and lighttpd also
- added an update sequence (instead of only modifying froxlor.sql)
- added english language variables
- added missing parts in admin_vhostsettings.php
- added parameter replacements as available since PR #244 

The solution has been tested working on different setups and should integrate / upgrade flawlessly.

### DIFF
--- a/admin_vhostsettings.php
+++ b/admin_vhostsettings.php
@@ -2,18 +2,18 @@
 
 /**
  * This file is part of the Froxlor project.
- * Copyright (c) 2003-2009 the SysCP Team (see authors).
- * Copyright (c) 2010 the Froxlor Team (see authors).
+ * Copyright (c) 2016 the Froxlor Team (see authors).
  *
  * For the full copyright and license information, please view the COPYING
  * file that was distributed with this source code. You can also view the
  * COPYING file online at http://files.froxlor.org/misc/COPYING.txt
  *
  * @copyright  (c) the authors
- * @author     Florian Lippert <flo@syscp.org> (2003-2009)
  * @author     Froxlor team <team@froxlor.org> (2010-)
  * @license    GPLv2 http://files.froxlor.org/misc/COPYING.txt
  * @package    Panel
+ *
+ * @since      0.9.35
  *
  */
 
--- a/admin_vhostsettings.php
+++ b/admin_vhostsettings.php
@@ -1,0 +1,232 @@
+<?php
+
+/**
+ * This file is part of the Froxlor project.
+ * Copyright (c) 2003-2009 the SysCP Team (see authors).
+ * Copyright (c) 2010 the Froxlor Team (see authors).
+ *
+ * For the full copyright and license information, please view the COPYING
+ * file that was distributed with this source code. You can also view the
+ * COPYING file online at http://files.froxlor.org/misc/COPYING.txt
+ *
+ * @copyright  (c) the authors
+ * @author     Florian Lippert <flo@syscp.org> (2003-2009)
+ * @author     Froxlor team <team@froxlor.org> (2010-)
+ * @license    GPLv2 http://files.froxlor.org/misc/COPYING.txt
+ * @package    Panel
+ *
+ */
+
+define('AREA', 'admin');
+require './lib/init.php';
+
+$id = 0;
+if (isset($_POST['id'])) {
+	$id = intval($_POST['id']);
+} elseif (isset($_GET['id'])) {
+	$id = intval($_GET['id']);
+}
+
+if ($page == 'overview') {
+	if ($action == '') {
+		$tablecontent = '';
+		$count = 0;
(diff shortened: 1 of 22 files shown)


d00p:
Instead of posting a PR every 2 hours, why don't you just join us on IRC and we talk about all your ideas and stuff...it's a bit stressful :)

admin_vhostsettings.php
```diff
@@ -0,0 +1,233 @@
+<?php
+
+/**
+ * This file is part of the Froxlor project.
+ * Copyright (c) 2003-2009 the SysCP Team (see authors).
```
d00p: new files do not have anything to do with SysCP

admin_vhostsettings.php
```diff
@@ -0,0 +1,233 @@
...
+ * Copyright (c) 2010 the Froxlor Team (see authors).
+ *
+ * For the full copyright and license information, please view the COPYING
+ * file that was distributed with this source code. You can also view the
+ * COPYING file online at http://files.froxlor.org/misc/COPYING.txt
+ *
+ * @copyright  (c) the authors
+ * @author     Florian Lippert <flo@syscp.org> (2003-2009)
```
d00p: new files do not have anything to do with SysCP

lng/english.lng.php
```diff
@@ -1969,4 +1969,21 @@
...
+$lng['admin']['vhostsettings']['vhostsettings'] = 'Predefined VHost configuration';
+$lng['admin']['vhostsettings']['novhostsettings'] = 'No predefined VHost configuration';
+$lng['admin']['vhostconfig']['socket_dir'] = 'Will be replaced with the corresponding Unix socket path, e.g. /var/run/domain-php-fpm.socket.';
+$lng['admin']['vhostconfig']['docroot'] = 'Will be replaced with the DocumentRoot.';
+$lng['admin']['vhostconfig']['ip'] = 'Will be replaced with the IP address of the Vhost.';
+$lng['admin']['vhostconfig']['port'] = 'Will be replaced with the port number of the Vhost.';
+$lng['admin']['vhostconfig']['scheme'] = 'Will be set to <strong>http</strong> or <strong>https</strong> in case SSL / TLS is activated for the Vhost.';
+$lng['admin']['vhostconfig']['vhost_usedefaultlocation']['title'] = 'Use the standard Vhost location directives';
+$lng['admin']['vhostconfig']['vhost_usedefaultlocation']['desc'] = 'If activated, the standard Vhost location directives, such as <br>location / { ... }<br> location @rewrites { ... }<br> location ~ ^(.+?\.php)(/.*)?$ { ... }<br> are excluded from the Vhost configuration file.';
```
d00p: nginx specific...

Churro: In this case, "such as" is used as a synonym for "for example". Of course, you could mention the default directives of all three webservers here but that would probably miss the point.

Churro:
Alright, Copyright fixed. Any other issues?

lng/english.lng.php
d00p: Well even "vhost location directives" is nginx specific....no apache-user knowns what a location-directive is (generally speaking). Something more abstract and "valid" for all webservers would be nice (imho)

Churro: While looking for a more generalized description, I noticed that it actually makes no sense to show this option to Apache / lighttpd users at all. Aside from nginx, no default diroptions are included / required anyway so it is somehow redundant to let the user disable it.

Summarizing, commit 373cb91 now simply hides this option for webservers other than nginx and sets it to the default value in the db (--> thereby, retaining the option that the admin may change the webserver occasionally). The commit also includes a typo fix and fixes the indentation of the SQL statement in the db upgrade step.

nachtgeist:
This will lead to serious confusion when the webserver is switched as this might lead to e.g. apache config directives ending up in ngingx server-configs.

The easiest way to fix this would be to provide separate textareas for each supported webserver and in the cron scripts only to incoroporate the contents from the appropriate column.

EleRas:
But this is already the case with the "default vhostsettings", they break as soon as you switch webservers

nachtgeist:
Good point, didn't realize that as I never used that myself.

Then they need be split up per webserver as well.

d00p:
I'd suggest to add a field 'Webserver' to the table which leads to a select box and then you can create templates for whatever Webserver and they are only used if the templates Webserver matches the currently used...

Churro:
Alright, this also matches with what I just proposed in IRC. A PR is going to follow soon.

PR #319, opened 2016-02-21
Add missing function argument

stweil:
The function "print_help" expects an argument.

Signed-off-by: Stefan Weil <sw@weilnetz.de>


d00p:
just a note: this script is ads good as unused :P

stweil:
What's the reason why it is rarely used? It reports a large number of issues in the translations.
By the way: did you ever think about replacing the current localization by gettext, for example?

d00p:
we do not speak all the languages, that's why we cannot update all of them. And yes, there were ideas etc.pp. for replacing. I won't start a discussion about that here. It has been discussed many times.

PR #323, opened 2016-02-23
Improve autoupdate against HTTP MITM and MD5 collisions

Churro:
**Problem:**
Currently, the autoupdater suffers from two vulnerabilities:

1. The Froxlor host can be tricked into updating with a non-official (or even backdoored) update archive. Using DNS spoofing an attacker is able to perform Cache Poisoning on the host running Froxlor. For an attacker, it is then easy to masquerade as autoupdate.froxlor.org since the connection is HTTP only.

2. It is trivial to produce an update archive which has the same MD5 hashsum as the official one. MD5 is known to be officially broken for more than 10 years now, it takes modern GPUs only < 30min to find a collision. Any modification to the update file would hence not be detectable. However, since the checksum is delivered using HTTP anyway, an attacker may deliver any hashsum.

**Solution**:
- Change HTTP to HTTPS. Therefore, you also have to roll-out the updates via HTTPS on the server-side. Also enable HSTS on autoupdate.froxlor.org in order to prevent an sslstrip attack.
- Prevent trivial collisions by relying on a currently more secure hash function, e.g. SHA-256

Enable HPKP on autoupdate.froxlor.org, store the fingerprint in Froxlor as a constant and verify it upon connecting to the host.

### DIFF
--- a/admin_autoupdate.php
+++ b/admin_autoupdate.php
@@ -22,9 +22,9 @@ define('AREA', 'admin');
 require './lib/init.php';
 
 // define update-uri
-define('UPDATE_URI', "http://version.froxlor.org/Froxlor/legacy/" . $version);
-define('RELEASE_URI', "http://autoupdate.froxlor.org/froxlor-{version}.zip");
-define('MD5SUM_URI', "http://autoupdate.froxlor.org/froxlor-{version}.zip.md5");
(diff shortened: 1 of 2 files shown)


d00p:
Can only be merged when server-side changes are done. Thx so far.

PR #325, opened 2016-02-27
ProFTPD: enable TLS by default. Fixes #1522

Churro:
From a security perspective, it seems adequate to enable TLS for FTP by default. This also fixes bug report #1522 which postulates settings that also comply with FileZilla.

### DIFF
--- a/lib/configfiles/gentoo.xml
+++ b/lib/configfiles/gentoo.xml
@@ -139,7 +139,7 @@ server.errorlog      = var.logdir  + "/error.log"
 
 server.indexfiles    = ("index.php", "index.html",
 						"index.htm", "default.htm")
-						
+
 server.name			 = "<SERVERNAME>"
 server.port          = 80
 server.bind          = "<SERVERIP>"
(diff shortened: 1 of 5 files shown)


d00p:
and what about gentoo and rhel_centos?

Churro:
Ok :P 
Added the updated template for Gentoo now. For CentOS, TLS with ProFTPd is already enabled in the current template.

Churro:
- Ehm... are you sure you know how the BREACH attack works?
Basically, by exploiting compression in the HTTP protocol of requested wrapped within the TLS connection. To be successful, you also need reflected user input in the response body. While there is data compression with FTP (MODE Z), I doubt that you can reflect input or a secret (e.g. CSRF token) within FTP responses :laughing: 

To be serious: BREACH is no attack on TLSv1 itself. Hence, there is currently no need to disable TLSv1. Practically, you might even exclude a wide set of clients in case they do not support > TLSv1.

- In terms of the cipher suites, I agree. Nevertheless, before specifying something explicit here, you should know what security libs (OpenSSL?) are used underneath the most common FTP clients. The problem is: By specifying something like ALL:!ADH:!RC4, you do not disable all insecure ciphers, like null ciphers, SEED, PSK, etc. On the other hand, if you narrow it down too much, at some point you also risk to block a set of ciphers, a major FTP client depends on ultimately. A compromise would be to simply exclude RC4 temporarily. The cleaner solution, of course, would be to restrict the set to an explicit list (whitelisting instead of blacklisting) of broadly supported but secure ciphers.

- I partly agree on making TLS mandatory by default. IMHO this should be the decision of an admin since it has a notable effect on all clients using the service. So, I assume an admin should explicitly be aware of the fact if FTP access is only possible having TLS enabled.

PR #327, opened 2016-02-28
Changes in the config templates

iandk:
Hello,

I updated the following things in the config themplates:

enabled plaintext auth because with no TLS enabled, you can´t login to the maildeamon
disabled RLimitChroot because it brings some problems with it in shared hosting


d00p:
three year old PR, sorry, close

PR #329, opened 2016-03-03
Set myhostname to $mydomain in postfix

HolySephi:
If you configure Froxlor properly to a valid FQDN like srv.domain.tld you would want to set myhostname in postfix to that FQDN and not as previously to mail.srv.domain.tld

### DIFF
--- a/lib/configfiles/gentoo.xml
+++ b/lib/configfiles/gentoo.xml
@@ -1061,11 +1061,11 @@ program_directory = /usr/libexec/postfix
 sendmail_path = /usr/sbin/sendmail
 
 ## General Postfix configuration
-# should be the default domain from your provider eg. "server100.provider.tld"
+# FQDN from Froxlor
 mydomain = <SERVERNAME>
 
-# should be different from $mydomain eg. "mail.$mydomain"
-myhostname = mail.$mydomain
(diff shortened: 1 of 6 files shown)


d00p:
did you miss gentoo?

Marco01809:
These comments:
`# should be different from $mydomain eg. "mail.$mydomain"`
should be removed then.

I always wondered why Froxlor was doing this and didn't simply use the hostname. Is there any reason to this?

HolySephi:
hrm i forked with the changes to the gentoo file - maybe github bugged out a bit - i'll try to add the file.

Marco: the reason is simple: this is the default postfix configuration of the distributions - it's designed to be used with a single domain and therefore to have "domain.tld" as receiving domain for local users like root@domain.tld. But to not break some RFC rules you have to use a FQDN as EHLO like mail.domain.tld or server.domain.tld.

But if you configure Froxlor properly you already use a FQDN as domain for the local users (like froxlor.domain.tld) because you want virtual users managed by froxlor for your @domain.tld

The comments are still from the default configuration because die initial template in Froxlor is just a copy from the default configuration with some additions to add the support for the Froxlor specific things - and setting myhostname to $mydomain should be one of these Froxlor specific settings.

PR #333, opened 2016-03-04
finalize reversion of PRs ##318+326: also revert DB changes

nachtgeist:
* drop table panel_vhostconfigs
* remove fields added to panel_domains
* increment version to -rc2
* white space fix


nachtgeist:
PS: sry for changing an -rc version, me not being a main dev and all, but going back to -dev seemed even more wrong ;)

d00p:
Thanks so far. Exactly because of the version change the needed DB updates are held back (i die not forget them) until the second RC

PR #334, opened 2016-03-06
ssl_cert_chainfile should only contain the intermediate certificate

RipClaw2971:
The column ssl_cert_chainfile in the domain_ssl_settings tablet should only contain the intermediate certificate. Using the fullcert from let's encrypt leads on using nginx to a certificate file like this:

[Certificate]
[Certificate]
[Intermediate Certificate]

Not a real problem but the check on ssllabs.com shows a error message:

[https://community.qualys.com/docs/DOC-1931](url)

### DIFF
--- a/scripts/jobs/cron_letsencrypt.php
+++ b/scripts/jobs/cron_letsencrypt.php
@@ -28,7 +28,7 @@ $certificates_stmt = Database::query("
 ");
 
 $updcert_stmt = Database::prepare("
-	REPLACE INTO `".TABLE_PANEL_DOMAIN_SSL_SETTINGS."` SET `id` = :id, `domainid` = :domainid, `ssl_cert_file` = :crt, `ssl_key_file` = :key, `ssl_ca_file` = :ca, `ssl_cert_chainfile` = :fullchain, `ssl_csr_file` = :csr, expirationdate = :expirationdate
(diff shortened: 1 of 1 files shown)


d00p:
but the CA is already $return['chain'], are you sure this is still correct afterwards?

RipClaw2971:
The certificate file for nginx ist created by the chain file and the certificate but not from the ca file:

`$dom_certs['ssl_cert_file'] = trim($dom_certs['ssl_cert_file'])."\n".trim($dom_certs['ssl_cert_chainfile'])."\n";`

And the configuration on Apache SSLCertificateChainFile uses the content from the field ssl_cert_chainfile.

RipClaw2971:
Here are two test results from SSLLabs:

This one is the original configuration with fullchain: [https://www.ssllabs.com/ssltest/analyze.html?d=cglobe.info](https://www.ssllabs.com/ssltest/analyze.html?d=cglobe.info)

This one is the one with only the chainfile in ssl_cert_chainfile: [https://www.ssllabs.com/ssltest/analyze.html?d=roundcube.cglobe.info](https://www.ssllabs.com/ssltest/analyze.html?d=roundcube.cglobe.info)

Take a look at "Chain issues"

d00p:
You mean this: Chain issues	Incorrect order, Extra certs?

RipClaw2971:
Yes

d00p:
why did you close it?

RipClaw2971:
Sorry mistake

PR #337, opened 2016-03-08
Fix for Bug #1603

RipClaw2971:
"Allow from all" makes more sense


lib/configfiles/wheezy.xml
```diff
@@ -71,7 +71,7 @@
 Alias "/.well-known/acme-challenge" "{{settings.system.letsencryptchallengepath}}/.well-known/acme-challenge"
 <Directory "{{settings.system.letsencryptchallengepath}}/.well-known/acme-challenge">
 	Order Deny,Allow
```
d00p: Then it should also be "Order allow, deny"

d00p:
Done, see https://github.com/Froxlor/Froxlor/commit/4330b08c0438ccaaa2ed31326e3c0a5273a00b2f

PR #338, opened 2016-03-09
Add DKIM pub key / DNS records to customer domain panel.

julianmatz:
Add DKIM information under Domains -> Edit in the customer panel.

fireshot screen capture 727 - netlink - froxlor server management panel

### DIFF
--- a/lib/functions/froxlor/function.generateDKIM.php
+++ b/lib/functions/froxlor/function.generateDKIM.php
@@ -1,0 +1,95 @@
+<?php
+
+	function generateDkim( $vars ) {
+
+		if ( Settings::Get( 'dkim.use_dkim' ) !== '1'
+			|| !isset( $vars['dkim_pubkey'] )
+			|| $vars['dkim_pubkey'] == '' ) {
+				return false;
+		}
+		
+		$selector = "custom";
+		
+		// start
+		$dkim_txt = 'v=DKIM1;';
+
+		// algorithm
+		$algorithm = explode( ',', Settings::Get( 'dkim.dkim_algorithm' ) );
+		$alg = '';
+
+		foreach ( $algorithm as $a ) {
+			if ( $a == 'all' ) {
+				break;
+			} else {
+				$alg .= $a . ':';
+			}
+		}
+
+		if ( $alg != '' ) {
+			$alg = substr( $alg, 0, -1 );
+			$dkim_txt .= 'h=' . $alg . ';';
+		}
+
+		// notes
+		if ( trim( Settings::Get( 'dkim.dkim_notes' ) != '' ) ) {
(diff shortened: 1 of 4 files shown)


d00p:
I think a fully DNS editor would make more sense

julianmatz:
Some hosting providers have external primary DNS servers, though. Or third party DNS services. Or customers might even decide to host their DNS records with their domain registrars. Like Dyn, ClouDNS, CloudFlare, etc. In this case, it would make sense to make the DKIM key available to copy and paste.

julianmatz:
@nachtgeist Sorry about that. My bad. I've amended the commit. The generateDKIM() function is called on line 710 in customer_domains.php.

lib/functions/froxlor/function.generateDKIM.php
```diff
@@ -0,0 +1,95 @@
...
+
+		// split if necessary
+		$dkim_record = '';
+		$lbr = 50;
+		for ( $pos = 0; $pos <= strlen( $dkim_txt ) -1; $pos += $lbr ) {
+			$dkim_record .= ( ( $pos == 0 ) ? '( "' : "\t\t\t\t\t \"" ) . substr( $dkim_txt, $pos, $lbr ) . ( ( $pos >= strlen( $dkim_txt ) - $lbr ) ? '" )' : '"' );
+		}
+
+		// Host entry
+		if ( !empty( $selector ) ) {
```
d00p: unnecessary, how can this ever be empty if you assign a value to $selector in line 11? Also, why is the selector hardcoded to "custom"?

d00p:
Why only do this for the customer? What about admin_domains?

julianmatz:
Hey, Michael. That was going to be my next step. I  will work on that. I just started with customer_domains because then, both admin and customer can get the record.

lib/functions/froxlor/function.generateDKIM.php
julianmatz: Thanks! I'll check on that. The "custom" selector doesn't have to be there. It just made sense to me because that's how I had it set up on my server - [servername].domainkey.[domain]. Maybe it could be a dynamic option under Settings -> DomainKeys?

d00p: It even should be a setting :)

d00p:
Well then I'll wait for this to be complete then

lib/functions/froxlor/function.generateDKIM.php
julianmatz: Agreed!

julianmatz:
OK, cool. Sounds good!

d00p:
as of https://github.com/Froxlor/Froxlor/tree/dns-editor, froxlor will soon be having a full qualified DNS editor so I think this PR is unnecessary

PR #339, opened 2016-03-09
Support for multi homed DNS servers

KZumbusch:
Hi,

the secondary DNS servers might have multiple IP addresses and the transfer requests might be issued from any of the IP addresses; therefore all IP addresses of the secodnary DNS servers should be allowed to transfer zones. I had this fix added to my old syscp installation for some years. 

As I plan to upgrade to a recent Froxlor version I decided to share some of my modifications with you.

Regards

Kay Zumbusch


d00p:
I'll fix that after merge

PR #342, opened 2016-03-27
Add the posibility to create new mailboxes with first- and lastname

virusbrain:
This patch provides the positbility to create a new mailbox with a firstname and a
lastname. This is very useful if you want to connect these mailboxes with a groupware.
For the case that the feature is not wanted: There was added a setting for this feature.
Default status of this feature is disabled.

### DIFF
--- a/actions/admin/settings/150.mail.php
+++ b/actions/admin/settings/150.mail.php
@@ -61,6 +61,14 @@ return array(
 					'string_emptyallowed' => true,
 					'save_method' => 'storeSettingField',
 					),
+				'panel_mailacc_with_name' => array(
+					'label' => $lng['serversettings']['mailacc_with_name'],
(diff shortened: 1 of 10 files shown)


d00p:
If wanted/needed why is the name not shown on the mail-address overview then? You can see whether a name was stored for an account / address only if you edit the address

virusbrain:
Because the name can only be shown if the mail-address is a real mailbox - not only a forwarding address. If the name should be shown I think it would be better to spilt the email part into two seperate pages: One Mailbox overview and one forward-address overview.

d00p:
Well, you know you can use if-statements in templates? I dont hink this makes any sense if you do not display the stored name anywhere...if theres no reaction here this PR will be closed

d00p:
no reaction for over a year, closed as said

PR #344, opened 2016-03-31
Plugin system

eis-os:
Hello

As requested on the forums, this is a rebase of my plugin system for review.

Small technical overview:

Events in FroxlorEventConstants can be fired via FroxlorEvent::eventname() shorthand.
References can be used if data needs to be passed back.

FroxlorPlugins loads the current active plugins if necessary and provides information if any plugin need an update. The plugin update ui currently is very basic and a bit bit buggy to allow easier merging. 
The formfield allows changing settings that aren't yet in the db. (this is an explicit option)

Settings for plugins are stored by using the pluginid as group. The FroxlorPlugin provides the necessary infrastructure, so it's shouldn't be necessary to call settings directly for them.


sebastian-schmid:
To which Topic is this related and is there an issue at https://redmine.froxlor.org/projects/froxlor for this?

d00p:
Why would that be relevant?

eis-os:
As example related to this:  https://redmine.froxlor.org/issues/21

Or this: 
https://redmine.froxlor.org/issues/702
There a lot stuff on the bugtracker that could theoretical benefit.

https://forum.froxlor.org/index.php/topic/13264-hook-system/

Froxlor will never be able to support all configuration wishes out of the box.
Technical it would be much easier to develop / modify some features without touching a lot internals if froxlor would be modular.

I am still not happy with it and the internal handling of stuff in Froxlor:
A good example is I wrote a plugin for creating additional letsencrypt certs for the panel/mailserver.
To reuse the lescript class I had to use reflection because the api assumes

sebastian-schmid:
@d00p I just want to read more about the plans on this topic

@eis-os thanks for the links and information

@d00p will you accept this merge request? Are there any further changes needed?

d00p:
Possibly. But why create a new logger and settings class? I understand that you want to wrap the settings for the specific plugin but it's kind of irritating and makes maintenance harder, also a couple of screenshots and a bit of documentation on howto plugins work and what they need to look like would be awsome

eis-os:
The intention for FroxlorPluginSettings to move settings into it's own settings table eventually.
It could be technical possible to overwrite a "namespace" with the right or better say wrong pluginname currently.

This should be details not known to a plugin. 
FroxlorPluginTemplate has the same intention, have the template system plugins separate to the Froxlor template system.

The current branch is written in a style to change as few lines of Froxlor as possible,
there are still a ton of stuff not handled correctly:

- Required minimum Froxlor version can't be set
- No way to mark plugin system version compatibility.
- Need to load plugin class to get version
- Can't split code (frontend/cron) into several files

## Logger Code: 
Currently all plugin logs are passed through class.FroxlorPlugins.php logPluginMessage.
The flow is like this: 
Plugin->log   =>    FroxlorPlugins::logPluginMessage()  =>  FroxlorPlugins->_logAction->log.
logAction can be a FroxlorLogger, a LoggerTee, FroxlorPluginsDebugLog.

Again this hides any implementation details of logging for the plugin. The frontend AREA difference as example. For cron runs the Tee+FroxlorPluginsDebugLog is responsible to hide the current global debugHandle detail.

### Log Rewrite
This currently create a lot glue code behind the scenes, in the long run it would be much better if the log systems could be changed to be more unified across the whole code base.
````
$cronlog->logAction(CRON_ACTION, LOG_WARNING, 'Automatic update done - you should check your settings to be sure everything is fine');
fwrite($debugHandler, '*** WARNING *** - Automatic update done - you should check your settings to be sure everything is fine' . "\n");
````

Would become:
````
FroxlorLog::log(LOG_WARNING, 'Automatic update done - you should check your settings to be sure everything is fine');
or 
FroxlorLog::warning('Automatic update done - you should check your settings to be sure everything is fine');
````

When cron runs FroxlorLog would hold a LoggerTee outputting to selected log method class + debugLog text file class (the current FroxlorPluginsDebugLog class)

After rewrite, you could simply change FroxlorPluginLog to be a simple proxy to pass all stuff to FroxlorLog without changing any code in plug-ins.

## Sample Plugin
https://github.com/eis-os/FroxlorPlugin-nssextrausers

PR #345, opened 2016-04-01
Update cron_letsencrypt.php

Amaridian:
When using data from previous cert, check if it was issued by let's encrypt to prevent invalid data.

### DIFF
--- a/scripts/jobs/cron_letsencrypt.php
+++ b/scripts/jobs/cron_letsencrypt.php
@@ -47,11 +47,16 @@ foreach($certrows AS $certrow) {
 	{
 		$cronlog->logAction(CRON_ACTION, LOG_DEBUG, "Updating " . $certrow['domain']);
 
+		// Extract old SAN-data
+		$x509data = false;
 		if ($certrow['ssl_cert_file']) {
+			$x509data = openssl_x509_parse($certrow['ssl_cert_file']);			
+		}
+		
+		// Check for existing Cert and Let's Encrypt as issuer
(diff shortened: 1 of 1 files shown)


d00p:
Tested?

nachtgeist:
What by your definition is invalid data?

Amaridian:
it is tested (and solved problems) on two of my servers. 

In my case, i had a StartSSL wildcard cert installed and activated letsencrypt for that domain. The froxlor cronjob took this wildcard-entry from my old cert to request a new cert from letsencrypt. This resulted in an error message from letsencrypt (no wildcard-certs allowed). 

The old code didnt check an existing cert for its issuer.... with my changes it does exactly that. Old cert data is only referenced when the issuer is letsencrypt.

nachtgeist:
While it makes sense what you're trying to accomplish, the textual issuer verification IMO is pointless. 

So unless you check against the LE CA certificates I vote against this.

Amaridian:
Well, generally, the whole import of any data from "old" certificates does not make a lot of sense. The cronjob should check for the current settings for the current domain, more or less calculate the needed subdomains and then check if all this data coresponds with the data from the last known cert. Based on this it should either renew the current cert or get a new one.

nachtgeist:
With PR #353 merged, this is no longer relevant. CSRs now get re-created every time a certificate is to be requested from LE, disregarding any pre-existing CSRs.

d00p:
Obsolete as of https://github.com/Froxlor/Froxlor/pull/353

PR #347, opened 2016-04-11
min-height against height fix for ajax-changing-content

floreno:


d00p:
A description would be helpful...

PR #348, opened 2016-04-11
Added option in system/ssl-settings to define the desired ssl_protocols to be used

dschallert:
This commit adds the opportunity to choose the desired ssl protocols within System --> SSL Settings.
The available options are TLSv1, TLSv1.1, TLSv1.2  - which can be enabled in combination or single.

Since i could not figure out the Lighttpd config yet, i only added support for Nginx and Apache for now.

! This commit needs further testing !

### DIFF
--- a/install/updates/froxlor/0.9/update_0.9.inc.php
+++ b/install/updates/froxlor/0.9/update_0.9.inc.php
@@ -2491,6 +2491,10 @@ if (isFroxlorVersion('0.9.30-rc1')) {
 	Database::query("INSERT INTO `panel_settings` SET `settinggroup` = 'system', `varname` = 'ssl_cipher_list', `value` = 'ECDHE-RSA-AES128-SHA256:AES128-GCM-SHA256:RC4:HIGH:!MD5:!aNULL:!EDH'");
 	lastStepStatus(0);
 
+        showUpdateStep("Adding ssl-protocols setting");
+        Database::query("INSERT INTO `panel_settings` SET `settinggroup` = 'system', `varname` = 'ssl_protocols', `value` = 'TLSv1,TLSv1.1,TLSv1.2'");
+        lastStepStatus(0);
+
 	updateToVersion('0.9.30');
 }
 
--- a/actions/admin/settings/131.ssl.php
+++ b/actions/admin/settings/131.ssl.php
@@ -31,6 +31,16 @@ return array(
 									'save_method' => 'storeSettingField',
 									'overview_option' => true
 							),
+                                                       'system_ssl_protocols' => array(
+                                                                        'label' => $lng['serversettings']['ssl']['ssl_protocols'],
+                                                                        'settinggroup' => 'system',
+                                                                        'varname' => 'ssl_protocols',
+                                                                        'type' => 'option',
+                                                                        'default' => 'TLSv1,TLSv1.1,TLSv1.2',
+                                                                        'option_mode' => 'multiple',
+                                                                        'option_options' => array('TLSv1' => 'TLSv1', 'TLSv1.1' => 'TLSv1.1', 'TLSv1.2' => 'TLSv1.2'),
+                                                                        'save_method' => 'storeSettingField',
+                                                        ),
 							'system_ssl_cipher_list' => array(
 									'label' => $lng['serversettings']['ssl']['ssl_cipher_list'],
 									'settinggroup' => 'system',
--- a/install/froxlor.sql
+++ b/install/froxlor.sql
@@ -493,6 +493,7 @@ INSERT INTO `panel_settings` (`settinggroup`, `varname`, `value`) VALUES
 	('system', 'mod_fcgid_defaultini_ownvhost', '2'),
 	('system', 'awstats_icons', '/usr/share/awstats/icon/'),
 	('system', 'ssl_cert_chainfile', ''),
+	('system', 'ssl_protocols', 'TLSv1,TLSv1.1,TLSv1.2'),
 	('system', 'ssl_cipher_list', 'ECDH+AESGCM:ECDH+AES256:!aNULL:!MD5:!DSS:!DH:!AES128'),
 	('system', 'nginx_php_backend', '127.0.0.1:8888'),
 	('system', 'perl_server', 'unix:/var/run/nginx/cgiwrap-dispatch.sock'),
--- a/lng/english.lng.php
+++ b/lng/english.lng.php
@@ -1730,6 +1730,8 @@ $lng['domains']['serveraliasoption_none'] = 'No alias';
 $lng['error']['givendirnotallowed'] = 'The given directory in field %s is not allowed.';
 $lng['serversettings']['ssl']['ssl_cipher_list']['title'] = 'Configure the allowed SSL ciphers';
 $lng['serversettings']['ssl']['ssl_cipher_list']['description'] = 'This is a list of ciphers that you want (or don\'t want) to use when talking SSL. For a list of ciphers and how to include/exclude them, see sections "CIPHER LIST FORMAT" and "CIPHER STRINGS" on <a href="http://openssl.org/docs/apps/ciphers.html">the man-page for ciphers</a>.<br /><br /><b>Default value is:</b><pre>ECDH+AESGCM:ECDH+AES256:!aNULL:!MD5:!DSS:!DH:!AES128</pre>';
+$lng['serversettings']['ssl']['ssl_protocols']['title'] = 'Configure the TLS protocol version';
+$lng['serversettings']['ssl']['ssl_protocols']['description'] = 'This is a list of ssl protocols that you want (or don\'t want) to use when talking SSL. <b>Notice:</b> Some older browsers may not support the newest protcol versions. <br /><br /><b>Default value is:</b><pre>TLSv1, TLSv1.1, TLSv1.2</pre>';
 
 // Added in Froxlor 0.9.31
 $lng['panel']['dashboard'] = 'Dashboard';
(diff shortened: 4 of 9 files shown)


install/updates/froxlor/0.9/update_0.9.inc.php
```diff
@@ -2491,6 +2491,10 @@
 	Database::query("INSERT INTO `panel_settings` SET `settinggroup` = 'system', `varname` = 'ssl_cipher_list', `value` = 'ECDHE-RSA-AES128-SHA256:AES128-GCM-SHA256:RC4:HIGH:!MD5:!aNULL:!EDH'");
 	lastStepStatus(0);
 
+        showUpdateStep("Adding ssl-protocols setting");
```
d00p: Wrong place for the update, way wrong...

dschallert: Where to add it then?

d00p: At the end of the file; also increase db_version in /lib/version.inc.php (YYYYMMDDC) and in install/froxlor.sql

Just check other Updates, but what you did is just for people <0.9.30

dschallert:
Guess it should be better now.

dschallert:
Should be better now :-)

install/updates/froxlor/0.9/update_0.9.inc.php
```diff
@@ -3292,3 +3292,12 @@
 
 	updateToVersion('0.9.35.1');
 }
+
+if (isDatabaseVersion('201603150')) {
+
+	showUpdateStep("Adding ssl-protocols setting");
+	Database::query("INSERT INTO `panel_settings` SET `settinggroup` = 'system', `varname` = 'ssl_protocols', `value` = 'TLSv1,TLSv1.1,TLSv1.2'");
```
d00p: Please use Settings::AddNew...

dschallert:
Done.

install/updates/froxlor/0.9/update_0.9.inc.php
dschallert: Done.

dschallert:
Done.

install/updates/froxlor/0.9/update_0.9.inc.php
```diff
@@ -3295,6 +3295,8 @@
 
 if (isDatabaseVersion('201603150')) {
 
+	Settings::AddNew("panel.db_version", "201604120");
```
d00p: I explicitly commented on the corresponding line in my last comment...that's not what i meant, you do not add "panel.db_version" as NEW setting....what i meant was you actual NEW SETTING which would be system.ssl_protocols, therefore please remove this line and exchange your INSERT INTO sql line below with

Settings::AddNew("system.ssl_protocols", "...");

Also, please check out "preconfig" stuff so the admin can set the initial value of this setting right from within the update, and not gets the hardcoded default

dschallert:
And again.. I hope it's ok now.

dschallert:
And again.. I hope it's ok now.

install/updates/froxlor/0.9/update_0.9.inc.php
```diff
@@ -3295,10 +3295,10 @@
 
 if (isDatabaseVersion('201603150')) {
 
-	Settings::AddNew("panel.db_version", "201604120");
+	Settings::AddNew("system.ssl_protocols", "TLSv1,TLSv1.1,TLSv1.2");
 	
-	showUpdateStep("Adding ssl-protocols setting");
-	Database::query("INSERT INTO `panel_settings` SET `settinggroup` = 'system', `varname` = 'ssl_protocols', `value` = 'TLSv1,TLSv1.1,TLSv1.2'");
+	showUpdateStep("Adding ssl_protocols setting");
+	Database::query("INSERT INTO `" . TABLE_PANEL_SETTINGS . "` SET `settinggroup` = 'system', `varname` = 'ssl_protocols', `value` = 'TLSv1,TLSv1.1,TLSv1.2'");
```
d00p: remove this line...and place the Settings::AddNew() here....the insert into is EXACTLY what Settings::AddNew() does

d00p:
still missing preconfig stuff

dschallert:
okay, any hints? thanks

d00p:
https://github.com/Froxlor/Froxlor/blob/master/install/updates/preconfig/0.9/preconfig_0.9.inc.php

dschallert:
@d00p , I see no point why this patch should have a pre-config?

The default value for ssl_protocols it's set to "TLSv1,TLSv1.1,TLSv1.2", which is exactly the same as the current default value in the webserver config cronjobs in 0.9.35.1.  The patch basically just adds another menu entry into the UI in order to choose the desired TLS protocol versions if the admin wants to change them.

Anyways most of the users probably won't even ever change this setting, because they still need compatibility with stone age browsers such as IE 7-9.

negrusti:
Any site that wants to be PCI DSS compliant needs to disable TLSv1, this option is needed

d00p:
@dschallert preconfig is a nice way of giving the admin the opportunity to 1) see new settings and 2) adjust them BEFORE any cronjob runs - so yes, I'd say it should be added here for the new settings

dschallert:
added the missing preconfig stuff. 
needs testing.

install/updates/preconfig/0.9/preconfig_0.9.inc.php
```diff
@@ -708,4 +708,16 @@ function parseAndOutputPreconfig(&$has_preconfig, &$return, $current_version, $c
 	    $question.= makeyesno('enable_letsencrypt', '1', '0', '1').'<br />';
 	    eval("\$return.=\"" . getTemplate("update/preconfigitem") . "\";");
 	}
+	
+	if(versionInUpdate($current_db_version, '201603150')) {
+	    $has_preconfig = true;
+	    $description  = 'You can chose which TLS versions will be enabled for your webservers SSL setup.<br />Please select the desired TLS versions below.<br /><br />';
+	    $question = '<strong>Do you want to enable Let\'s Encrypt? (default: yes):</strong>&nbsp;';
+	    $question .= '<select name="update_ssl_protocols">';
+	    $options = makeoption('TLSv1', 'TLSv1', 'TLSv1');
```
d00p: The default select is not a "multiple" select, so setting each option as selected would end up in only one option being selected by default.

This means the select should look like this:
`<select name="update_ssl_protocols[]" multiple="multiple">`

which then leads to a different handling of the POST-value in https://github.com/Froxlor/Froxlor/pull/348/files#diff-29773a1587adda5af698b968f2130e49R3301 (it will be an array, not just a string)

Also, you are inserting the new setting using Settings::AddNew in https://github.com/Froxlor/Froxlor/pull/348/files#diff-29773a1587adda5af698b968f2130e49R3298 but then SQL-Insert the values from the preconfig manually AGAIN to the settings (without using the Settings::AddNew() function)  @ https://github.com/Froxlor/Froxlor/pull/348/files#diff-29773a1587adda5af698b968f2130e49R3302 - please fix (just look and earlier code how we do it)

install/updates/preconfig/0.9/preconfig_0.9.inc.php
```diff
@@ -708,4 +708,12 @@ function parseAndOutputPreconfig(&$has_preconfig, &$return, $current_version, $c
 	    $question.= makeyesno('enable_letsencrypt', '1', '0', '1').'<br />';
 	    eval("\$return.=\"" . getTemplate("update/preconfigitem") . "\";");
 	}
+	
+	if(versionInUpdate($current_db_version, '201603150')) {
+	    $has_preconfig = true;
+	    $description  = 'You can chose which TLS versions will be enabled for all your webservers SSL vhosts.<br /><br />';
+	    $question = '<strong>Please select the desired TLS versions below (default: TLSv1,TLSv1.1,TLSv1.2):</strong>&nbsp;';
+	    $question .= '<input type="text" id="update_ssl_protocols" value="TLSv1,TLSv1.1,TLSv1.2"/>'
```
d00p: why make this a text-field now when the setting is a multiple-select-box? People can enter values as they like which will not be available for selection in the settings -> no good

install/updates/froxlor/0.9/update_0.9.inc.php
```diff
@@ -3292,3 +3292,15 @@
 
 	updateToVersion('0.9.35.1');
 }
+
+if (isDatabaseVersion('201603150')) {
+
+	Settings::AddNew("system.ssl_protocols", "TLSv1,TLSv1.1,TLSv1.2");
```
d00p: remove this, it's wrong

install/updates/froxlor/0.9/update_0.9.inc.php
```diff
@@ -3298,7 +3298,7 @@
 	Settings::AddNew("system.ssl_protocols", "TLSv1,TLSv1.1,TLSv1.2");
 	
 	showUpdateStep("Adding ssl_protocols setting");
-	$update_ssl_protocols = isset($_POST['update_ssl_protocols']) ? $_POST['update_ssl_protocols'] : "TLSv1,TLSv1.1,TLSv1.2";
+	$update_ssl_protocols = isset($_POST['update_ssl_protocols']) ? implode(',', $_POST['update_ssl_protocols']) : "TLSv1,TLSv1.1,TLSv1.2";
 	Database::query("INSERT INTO `" . TABLE_PANEL_SETTINGS . "` SET `settinggroup` = 'system', `varname` = 'ssl_protocols', `value` = '" . $update_ssl_protocols ."'");
```
d00p: as said many times, this should be
`Settings::AddNew("system.ssl_protocols", $update_ssl_protocols);`

Tarry91:
Is there any progress here? I would very much like the possibility to change the SSL Protocols used.

d00p:
Integrated, see https://github.com/Froxlor/Froxlor/tree/4bcdfc0786067b69f9eef1c3da60d51ed2c625be

PR #349, opened 2016-04-13
Letsencrypt alias domains 

eis-os:
At least for Apache:

ServerAlias is used for ssl alias domains, however the certificate is only for the primary domain.
This patch adds the alias domains as additional san domains.  The primary domain currently is responsible for activating let's encrypt. 

This won't solve the problem that changing wwwalias settings or alias domains won't recreate a proper new certificate request as the old domains will always used.
Technical expire needs to be reset and the csr recreated when csr old domain list != domains.

Patch 2 $csrfile:
Seems like a variable from upstream lescript was not converted to database fields (untested)

### DIFF
--- a/lib/classes/ssl/class.lescript.php
+++ b/lib/classes/ssl/class.lescript.php
@@ -218,7 +218,7 @@ class lescript
 
 		$this->client->getLastLinks();
 
-		if (empty($csrfile) || Settings::Get('system.letsencryptreuseold') == 0) {
+		if (empty($csr) || Settings::Get('system.letsencryptreuseold') == 0) {
 			$csr = $this->generateCSR($privateDomainKey, $domains);
 		}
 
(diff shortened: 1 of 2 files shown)


d00p:
Tested?

eis-os:
I tested 716aab6 with one alias domain. 

- Removed the csr and set expire to null in TABLE_PANEL_DOMAIN_SSL_SETTINGS for the parent domain by hand
- In panel, enabled letsencrypt for the main domain. (no need to enable in alias, it won't work anyway)
- Regenerated via letsencrypt cron run + cron force

This worked fine. Domain + Alias domain works with www and without.

But how do you want to handle changes for wwwalias in general? 
cron_letsencrypt.php always uses the old csr domain data regardless of changes in the panel. (this behaviour isn't changed) 

The csr/csrfile change is untested.

nachtgeist:
I have https://redmine.froxlor.org/issues/1597 queued up. Will have a look at it over the weekend.

nachtgeist:
Sorry for the huge delay.

This PR can be closed, now implemented in #353 (thanks for your work though, eis-os :-) )

Basically your idea was correct.

I removed the code for recycling old CSRs. Didn't make much sense in the first place.

As for your questing regarding triggering CSRs: take a look at admin_domains.php from #353.

d00p:
Obsolete as of https://github.com/Froxlor/Froxlor/pull/353

PR #350, opened 2016-04-20
Prevent Chromes password autofill in

floreno:
If editing a customers in Chrome-Browser a saved password is automatically filled in the password field.
If not manually emptied the auto-filled-in-password gets the new customer-password.
This patch inserts a hidden password-field that gets auto-filled-in by chrome so the froxlor password-field keeps empty until filled manually.

### DIFF
--- a/templates/Sparkle/admin/customers/customers_edit.tpl
+++ b/templates/Sparkle/admin/customers/customers_edit.tpl
@@ -15,6 +15,7 @@ $header
 				<input type="hidden" name="action" value="$action" />
 				<input type="hidden" name="id" value="$id" />
 				<input type="hidden" name="send" value="send" />
+				<input type="password" name="fakepwd" class="input-fake" />
 			
 				<table class="full">
 					{$customer_edit_form}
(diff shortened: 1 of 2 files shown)


liwakura:
`<form autocomplete="false">` might be an simpler way to tell chrome not to autofill fields.

floreno:
@liwakura From the specs it should. But; try a google-search for "chrome ignores autocomplete"

liwakura:
See [this](http://stackoverflow.com/a/29582380) and [this](https://bugs.chromium.org/p/chromium/issues/detail?id=370363#c15).

BNoiZe:
Adding a field for chrome to autofill the password (even if it's hidden) is not secure, as the field will always be sent to the server and therefore it will leave your computer.

autocomplete="off" would be the only right way.

floreno:
@liwakura Sry my fault. I am working with a beta-version of chrome where the explained technique not works. Trying stable Chrome works like a charm :)
@BNoiZe Absolutely right.
Sorry for the inconvenience

PR #352, opened 2016-04-30
fix group by clause to be be compliant with mysql 5.7

FliegenKLATSCH:
In mysql 5.7 only_full_group_by is enabled by default.

### DIFF
--- a/lib/classes/idna/class.idna_convert_wrapper.php
+++ b/lib/classes/idna/class.idna_convert_wrapper.php
@@ -67,6 +67,16 @@ class idna_convert_wrapper
 		}
 	}
 
+	public function encode_uri($to_encode)
+	{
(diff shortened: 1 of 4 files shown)


d00p:
Are you going to find even more "fixes" or is this mergable (and tested?)

FliegenKLATSCH:
I don't know, if there are more of this issue.. I don't see any, but I'm not a "power-user"..
So you can merge it, it's tested.. If I find another one, I'll create another pull-request.

FliegenKLATSCH:
Whats the status now? Should I change something?

nachtgeist:
* please clean up the history of your pull request (i.e. fix-up ad6b39a and 779e41e into a single commit)
* you changed froxlor.sql - this file is used for a clean installation only. What's missing is code for migrating existing installations. Have a look at install/updates/froxlor/0.9/update_0.9.inc.php

PR #353, opened 2016-05-15
LE

nachtgeist:


nachtgeist:
- fixed botched code formatting
- add triggering of CSRs for domain editing by customers

PR #358, opened 2016-06-04
[SECURITY] Information disclosure on database failures

ohader:
In case the database is not responding, e.g. due to "too many connections"
cut-off database credentials might be shown and system path be revealed.
In terms of security this is considered as information disclosure.


ohader:
froxlor

The screenshot shows how that looked on my site without the change. Albeit the password was cut-off, it showed enough credentials already for a brute-force attack.

d00p:
Very nice, thx

PR #362, opened 2016-06-19
don't split .gitignore

nachtgeist:
they would end up in .deb packages


nachtgeist:
force-pushed my master: PR now includes SAPI checks

PR #363, opened 2016-06-19
some housekeeping in Debian packaging

nachtgeist:


nachtgeist:
update: replace courier by dovecot as the default imap/pop3 daemon

nachtgeist:
update: do NOT bump wheezy package's Standards field to 3.9.6

PR #364, opened 2016-06-20
Letsencrypt: Don't redirect .well-known

possi:
A redirect to foreign Domains or (not yet signed) https prevents letsencrypt-verification. So we should exclude .well-known/acme-challenge from redirects.


d00p:
apache only...

nachtgeist:
nothing to do with cronjob code - this is handled via config templates (for apache see /etc/apache2/conf-enabled/acme.conf)

PR #368, opened 2016-07-27
If a customer's page is hidden from menu via settings, redirect the u…

micw:
…ser if he tries to access the page


d00p:
Please remove all the superfluous exit-statements after the redirectTo() call as redirectTo() already calls exit at the end

PR #373, opened 2016-08-15
pdns record and zone creation improvements

BeryJu:
- only clear domains where account is froxlor or nothing (nothing since previous versions didn't set anything in the account column)
- create new zones with account 'froxlor' and type MASTER so they replicate to other nameservers

I am adding the account column so we can have domains that are not managed by froxlor in the PDNS database, without froxlor deleting them. This also makes it easier to delete froxlor domains from PDNS when they are removed from froxlor.

### DIFF
--- a/scripts/jobs/cron_tasks.inc.dns.20.pdns.php
+++ b/scripts/jobs/cron_tasks.inc.dns.20.pdns.php
@@ -1,6 +1,6 @@
 <?php
 if (! defined('MASTER_CRONJOB'))
-        die('You cannot access this file directly!');
+	die('You cannot access this file directly!');
 
 /**
  * This file is part of the Froxlor project.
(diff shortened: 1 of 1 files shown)


scripts/jobs/cron_tasks.inc.dns.20.pdns.php
```diff
@@ -104,23 +104,28 @@ private function _clearZoneTables($domains = null)
...
+				$this->_logger->logAction(CRON_ACTION, LOG_INFO, 'Ignored domain '.$domain['domain'].' because we do not manage it');
+			}
 		}
 	}
 
 	private function _insertZone($domainname, $serial = 0)
 	{
 		$ins_stmt = $this->pdns_db->prepare("
-			INSERT INTO domains set `name` = :domainname, `notified_serial` = :serial, `type` = 'NATIVE'
+			INSERT INTO domains set `name` = :domainname, `type` = 'MASTER', `account` = 'froxlor'
```
d00p: why did you remove the notified_serial?

BeryJu: Right, should've probably added that to the message.
I removed it since that field is used by PDNS to remember which serial the slaves were last notified with. By manually changing it, PDNS doesn't know that there is a difference between the local serial and the serial of the slave server, so it wouldn't send out notifications to the slaves.

d00p: OK, fair enough :+1:

d00p:
Did you test thoroughly?

BeryJu:
Tested it briefly, I'll get back to you when I'm done with some more thorough testing.

d00p:
awesome, would like to merge that for 0.9.37.1 (or 0.9.38 depending)

d00p:
Your recent changes are weird...that way, no domains that were deleted from froxlor is being removed from PDNS, your first solution did that using the account = 'froxlor' ....instead of iterating through the $domains array (like i do this now) you could simply DELETE ... WHERE account = 'froxlor' - what do you think?

BeryJu:
I am aware that the code is overcomplicated. We just had to get it working for a client yesterday, and I didn't want to leave barely-tested code here. I am going to take another look at it today/tomorrow.

My main goal with this is to modify the PDNS Interface so that it doesn't recreate all the domains/records/metadata every time, but rather just updates them.

d00p:
No more work by OP in half a year, close

PR #376, opened 2016-08-28
Set a User Agent in the HTTP request fetching the LE challenge URI

w6g23:
... while doing the self check.

A rule (e.g. Wordpress plugin iThemes) might block requests with empty
User Agents.

### DIFF
--- a/lib/classes/ssl/class.lescript.php
+++ b/lib/classes/ssl/class.lescript.php
@@ -176,7 +176,9 @@ class lescript
 			$this->log("Token for $domain saved at $tokenPath and should be available at $uri");
 
 			// simple self check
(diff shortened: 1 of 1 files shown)


d00p:
Could you set the UserAgent to Froxlor/{version} like here: https://github.com/Froxlor/Froxlor/blob/master/lib/ajax.php#L60

nachtgeist:
Why not move this to a separate function, say, getUserAgentStr()?

d00p:
because until now it was only in one file...also, i think it's unnecessary to create a function for that

d00p:
I'll just do it myself

w6g23:
Thanks for taking over from here -- I was merely hunting down the 403.

PR #377, opened 2016-08-29
do not die after token self check

oliverrahner:
We have to finish the challenge request so that the auth does not linger in state "pending", but goes to "invalid". See https://forum.froxlor.org/index.php/topic/13463-lets-encrypt-zertifikate-werden-nicht-erneuert/#entry32895


d00p:
sounds legit

oliverrahner:
Quickest PR ever :+1:

d00p:
krkrkr, you should join us on IRC and hang out with us if you got a bunch of fixes and it seems you are testing well :)

PR #378, opened 2016-08-31
Remove dependency on hard coded agreement URL for Let's Encrypt

oliverrahner:
Change the process to first create a new registration, which delivers the current TOS url in the response's header, then modify the newly created registration to accept the agreement.

### DIFF
--- a/lib/classes/ssl/class.lescript.php
+++ b/lib/classes/ssl/class.lescript.php
@@ -30,7 +30,7 @@ class lescript
 {
 
 	// https://letsencrypt.org/repository/
-	public $license = 'https://letsencrypt.org/documents/LE-SA-v1.1.1-August-1-2016.pdf';
+	public $license;
 
 	private $logger;
 
@@ -79,8 +79,9 @@ class lescript
 			if ($this->client->getLastCode() != 201) {
 				throw new \RuntimeException("Account not initialized, probably due to rate limiting. Whole response: " . json_encode($response));
 			}
-
-			$this->postNewReg();
+			$this->license = $this->client->getAgreementURL();
+			
+			$this->postRegAgreement(parse_url($this->client->getLastLocation(), PHP_URL_PATH));
 			$this->log('New account certificate registered');
 		} else {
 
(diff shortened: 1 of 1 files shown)


d00p:
very nice, is this tested?

oliverrahner:
Yes.
Plus it relies on documented behaviour of the ACME API (https://letsencrypt.github.io/acme-spec/#rfc.section.6.3.p.5) so it should be stable for the future.

lib/classes/ssl/class.lescript.php
```diff
@@ -517,6 +528,13 @@ public function getLastLinks()
 		preg_match_all('~Link: <(.+)>;rel="up"~', $this->lastHeader, $matches);
 		return $matches[1];
 	}
+
+	public function getAgreementURL()
+	{
+		preg_match_all('~Link: <(.+)>;rel="terms-of-service"~', $this->lastHeader, $matches);
+		return $matches[1][0];
```
discordier: > If the server wishes to present the client with terms under which the ACME service is to be used, it MUST indicate the URI where such terms can be accessed in a Link header with link relation “terms-of-service”. As noted above, the client may indicate its agreement with these terms by updating its registration to include the “agreement” field, with the terms URI as its value.

IMO this means, the URI is optional (in theory, unlikely to encountet this with letsencrypt.org though).
Therefore the matches might not be populated here.

If the URI is not contained, it should also not be set in `postRegAgreement()` above.

I do not know letsencrypt indepth, this is just a concern that arose from reading the specs.

7hunderbug:
I've just been trying to renew my LE certs after updating to Froxlor 0.9.37 today (first time use of LE functionality in Froxlor for me) and discovered that it is entirely broken due to the hard-coded Agreement URL. I only discovered this after finding out that the cronjob console debug output was broken and then finding the fix by adding the "json_encode" in line 80 from commit https://github.com/Froxlor/Froxlor/commit/cd806b19f7019981e4e53dfc4a9411d8d69819a0 to my install.

I can confirm that with this PR and with the commit mentioned above, requesting LE certs now works. However, I did have to remove all old SSL config from the Froxlor database, turn SSL off and on again and I had to manually disable SSL redirect until the certs had arrived.

lib/classes/ssl/class.lescript.php
oliverrahner: You are absolutely right... Solution would be to add `if (!empty($this->license))` before sending the reg update. I wanted to change this before @d00p merged, but too late now :) Maybe on for a new PR.

d00p:
whoops :P Well then let's go for another PR :) I wanted to release 0.9.37.1 yesterday and was waiting for your changes but i didn't have the time after all - Now i'm off to holiday for two weeks - too bad

PR #380, opened 2016-09-09
Make path to acme.conf configurable

w6g23:
When using FreeBSD -- as I do --, trees /usr and /etc are for the "base system", and additionally installed software goes into /usr/local/. A suiting path for the acme.conf file therefore is /usr/local/etc/nginx/acme.conf (and not /etc/nginx/acme.conf). In order to make this possible, make the path to this acme.conf file configurable in Settings/SSL Settings.

As a workaround, I have used a symlink up to now.

I suspect the (/usr/local/) /etc/nginx/acme.conf file is only used whit nginx as the web server; as far as I have been able to see, these configuration settings are stored differently when using Apache for instance.

Also, an update script is missing from this pull request because I have no idea how and if at all I should add one. A new configuration entry "system.letsencryptacmeconf" would have to be entered into panel_settings.

### DIFF
--- a/actions/admin/settings/131.ssl.php
+++ b/actions/admin/settings/131.ssl.php
@@ -105,6 +105,15 @@ return array(
 								'default' => false,
 								'save_method' => 'storeSettingField'
 							),
+							'system_letsencryptacmeconf' => array(
+									'label' => $lng['serversettings']['letsencryptacmeconf'],
+									'settinggroup' => 'system',
+									'varname' => 'letsencryptacmeconf',
+									'type' => 'string',
+									'string_emptyallowed' => false,
+									'default' => '/etc/acme.conf',
+									'save_method' => 'storeSettingField',
+							),
 							'system_letsencryptca' => array(
 									'label' => $lng['serversettings']['letsencryptca'],
 									'settinggroup' => 'system',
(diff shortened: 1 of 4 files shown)


d00p:
1) there is no official FreeBSD support (no config  templates) so FreeBSD specific needs are not relevant

2) the pull-request does only half a Job if at all...also i dont see the need to customize this path...

w6g23:
Ad 2)
Well. Excuse me please if I don't know the 107.000 LOC application by heart. I tend to think that having a file path hard-coded somewhere in a source code file is generally a thing one would love to get rid of. Regardless of the OS used, I was proposing a way to do so. Pretty sure this is only the first half of what needs to be done, but anyway, it might be a starting point; even if it were only a starting point for a discussion of the issue.

Ad 1)
Is it *really* FreeBSD-specific to think that some application software (e.g. nginx) might be installed in /usr/local, or /opt, or ... you name it (instead of /usr)? When you say, FreeBSD-specific needs would not be relevant, do you really think you would support some of us trying[1]? 

[1] See https://forum.froxlor.org/index.php/topic/13196-support-for-freebsd/

d00p:
The different file path is only necessary (if at all) on FreeBSD. So as long as there are no FreeBSD templates and froxlor not officially supporting FreeBSD this is just not needed.

If the pull-request was complete, we can surely make this an option, but remember that this option value has to be used in all the config-templates etc.pp.

w6g23:
I see.

Of course it is not *needed*. I would consider it a step towards more independence from assumptions.

And: There is help needed, from you or from someone else: What would make this patch "complete"? What hides behind "etc. pp."? Alternatively, where can I read it up?

d00p:
I'll describe everything as soon as i'm back from my vacation.

d00p:
Let's try this quickly:

- new settings need to be entered in install/froxlor.sql for new installations
- in actions/admin/settings/131.ssl.php please make the remove 'string_emptyallowed' and replace with 'string_type' => 'file'
- you do not need the extra sentence "ATTENTION: Let's Encrypt is still in beta" for this setting, as it is just a webserver Alias
- you are missing the usage of the newly added setting in both, apache and lighttpd cronjobs
- you need to add the new setting to the system for updaters in the file updates/froxlor/0.9/update_0.9.inc.php  (see former parts on how this works)
- if you added your database-update stuff to updates/froxlor/0.9/update_0.9.inc.php you need to increase the dbversion in lib/version.inc.php, install/froxlor.sql and don't forget to set it in updates/froxlor/0.9/update_0.9.inc.php [updateToDbVersion()]

d00p:
You know what, I'll just do it. Maybe you and others are willing to participate in the freebsd.xml config-template file then?

d00p:
fixed in 4229d8dda4b24b1ba077a602e4aa8420f9e4e521

PR #390, opened 2016-09-24
fix group by clause to be be compliant with mysql 5.7

FliegenKLATSCH:
In mysql 5.7 only_full_group_by is enabled by default.

### DIFF
--- a/install/froxlor.sql
+++ b/install/froxlor.sql
@@ -66,7 +66,7 @@ CREATE TABLE `mail_virtual` (
   `id` int(11) NOT NULL auto_increment,
   `email` varchar(255) NOT NULL default '',
   `email_full` varchar(255) NOT NULL default '',
-  `destination` text NOT NULL,
+  `destination` text NOT NULL default '',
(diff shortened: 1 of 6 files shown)


install/froxlor.sql
```diff
@@ -66,7 +66,7 @@ CREATE TABLE `mail_virtual` (
   `id` int(11) NOT NULL auto_increment,
   `email` varchar(255) NOT NULL default '',
   `email_full` varchar(255) NOT NULL default '',
-  `destination` text NOT NULL,
+  `destination` text DEFAULT NULL,
```
d00p: you are absolutely right

FliegenKLATSCH:
Hmm.. I think I got an error, because there was no default value..
So I will change the default from null to empty string? I don't see the point here, but I will change it, probably tomorrow..

nachtgeist:
> Hmm.. I think I got an error, because there was no default value..

Ah I see.

> So I will change the default from null to empty string?

ACK

FliegenKLATSCH:
Done.. Anything else?

PR #392, opened 2016-09-26
Prevent duplicates in ipsandports table

negrusti:
Hi,
I manipulate Froxlor tables from my own scripts and think that this constraint will be right in this table.

### DIFF
--- a/install/froxlor.sql
+++ b/install/froxlor.sql
@@ -280,7 +280,8 @@ CREATE TABLE `panel_ipsandports` (
   `default_vhostconf_domain` text,
   `ssl_cert_chainfile` varchar(255) NOT NULL,
   `docroot` varchar(255) NOT NULL default '',
-  PRIMARY KEY  (`id`)
+  PRIMARY KEY  (`id`),
+  UNIQUE KEY `ip_port` (`ip`,`port`)
 ) ENGINE=MyISAM CHARSET=utf8 COLLATE=utf8_general_ci;
 
 
--- a/install/updates/froxlor/0.9/update_0.9.inc.php
+++ b/install/updates/froxlor/0.9/update_0.9.inc.php
@@ -3321,6 +3321,14 @@ if (isFroxlorVersion('0.9.35.1')) {
 	updateToVersion('0.9.36');
 }
 
+if (isFroxlorVersion('0.9.37')) {
+	showUpdateStep("Updating from 0.9.37 to 0.9.38", false);
+	showUpdateStep("Adding unique key to ipsandports table");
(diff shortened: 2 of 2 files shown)


negrusti:
Sorry for noob question - is my 2nd commit visible now in pull request? Also unsure about Froxlor version check

nachtgeist:
> is my 2nd commit visible now in pull request?

yes

> Also unsure about Froxlor version check

Since DB versioning was introduced you need to check for the database version instead... just add your migration code at the very end of that file and use `isDatabaseVersion('YYYYMMDD#')` with the value used in `updateToDbVersion()` in the previous migration.

And then there's another issue: this code will fail if there already are duplicate entries.
@d00p: any idea how to handle this in a sensible and useful way?

negrusti:
In my experience when duplicates are accidentally created, Froxlor generates unworkable vhost configurations and Apache does not start with it.

nachtgeist:
ok good enough for me

d00p:
1) froxlor does check for duplicates when adding a new IP/port, see https://github.com/Froxlor/Froxlor/blob/master/admin_ipsandports.php#L216 and following
So, i'm unsure where you get these...

2) totally misplaced update queries, as  @nachtgeist said, add your migration/update code at the end of the update-file and just increase database-version

negrusti:
1) All my fault
2) Moved update queries to the end of the update file with DB version increase, please check

d00p:
Sorry, forgot to mention, please also set $dbversion in lib/version.inc.php :)

negrusti:
Perhaps the code starting here:
https://github.com/Froxlor/Froxlor/blob/master/admin_ipsandports.php#L216
can be simplified now, with INSERT converted to REPLACE on line 226?

d00p:
@negrusti i prefer the way it is with output of a message: hey, this ip/port already exists. and not just "overwriting" - you never know if it was intended or by mistake

d00p:
Froxlor already checks for duplicates and outputs a corresponding error-message. Just adding an UNIQUE-KEY does not change anything about this

negrusti:
The problem is - at the moment Froxlor is poorly suited for automation. Why would you reject this change? Does it have a potential to break anything?

d00p:
Ah NOW someone answers...i should close PR's earlier so people react

negrusti:
I am sorry - I guess I missed some message earlier - I assumed it was already merged

d00p:
And no, it does not break anything but also does not improve anything really...so it's quit unnecessary, but hey, here you go, have fun :)

negrusti:
thanks! :)

PR #393, opened 2016-10-04
Enable Plaintext Authentification

ianklemm:
Sice we have no other authentification method enabled, or SSL setup, we can´t connect to the MDA without enabling plaintext authentification.


nachtgeist:
Pushing plaintext authentication data over an unencrypted link is reckless unless you're in a tightly controlled (v)LAN environment where you can consider ALL network components to be secure/uncompromised etc. This does NOT apply to...roughly...99.9% of our users, which are using Froxlor on a publicly available machine.

Froxlor's config templates are by no means binding, a sysadmin knowing what she's doing always can tweak them to her local needs.

PR #395, opened 2016-11-06
Option to add a prefix to systems users/groups needed for fcgid+suexec

dphi:
Accidental username& group collisions of froxlor administered users and shell users can be avoided using this feature.

### DIFF
--- a/actions/admin/settings/135.fcgid.php
+++ b/actions/admin/settings/135.fcgid.php
@@ -105,6 +105,14 @@ return array(
 					'default' => 30,
 					'save_method' => 'storeSettingField'
 					),
+				'system_mod_fcgid_system_prefix' => array(
+					'label' => $lng['serversettings']['mod_fcgid']['username_prefix'],
+					'settinggroup' => 'system',
+					'varname' => 'mod_fcgid_system_prefix',
+					'type' => 'string',
+					'default' => '',
+					'save_method' => 'storeSettingField'
+					),
 				)
 			)
 		)
--- a/install/froxlor.sql
+++ b/install/froxlor.sql
@@ -450,6 +450,7 @@ INSERT INTO `panel_settings` (`settinggroup`, `varname`, `value`) VALUES
 	('system', 'last_archive_run', '000000'),
 	('system', 'mod_fcgid_configdir', '/var/www/php-fcgi-scripts'),
 	('system', 'mod_fcgid_tmpdir', '/var/customers/tmp'),
+	('system', 'mod_fcgid_system_prefix', ''),
 	('system', 'ssl_cert_file', '/etc/apache2/apache2.pem'),
 	('system', 'use_ssl', '0'),
 	('system', 'default_vhostconf', ''),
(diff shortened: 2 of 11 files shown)


d00p:
eh, i don't quite get what you want to achieve here...?! What name collision?

Guite:
I think he refers to the edge case that a customer is added which is equally named as an already existing system user.
These changes make the queries more complex unfortunately though.

d00p:
@Guite if you specify custom usernames, they are checked against existing system user (see https://github.com/Froxlor/Froxlor/blob/master/admin_customers.php#L593) - if you leave the field empty, the defined prefix + incremented account number is used, it is not checked whether this is a system user already....but this pull-request is in no way a solution for this

d00p:
(though i don't get why one would add a systemuser named "web123" if that is the froxlor naming, but hey...people are creative sometimes :P)

mwalle:
The idea was, you don't want to pollute the user and group names namespace of the system, because you have no control over the username a reseller choose. So by prepending a fixed value you can at least make sure it won't collide with a (system) username you may want to have in the future. Also, it is possible to distinguish between system and froxlor users easily. And yes, that CONCAT(..) is quite ugly... In contrast to forcing a username to begin with a prefix, this solution is transparent for the user which still has the username "myname" but within the operating system the user is known as "prefixmyuser".

d00p:
well...if it's just the resellers you want to have username-control over, why not add an option for reseller-prefixes instead? I don't like this "hidden prefix" added to the username....What if you enabled ssh for one if your customers....he thinks his username is "web1", but instead it's "froxlor_web1" or something...not very nice

mwalle:
and yes, it doesn't work with the custom shell feature. But until now, I didn't even know there was such a feature (please note, that we had this patch for ages until @dphi tried to upstream it, which I guess won't happen ;) )

Nevertheless, I still think it is a bad idea to give users (resellers) control about what user exists on the system. What if a reseller creates a user that collides with a name of a system user (read user for a system service). Eg. reseller creates a user "stunnel4", what happens if you do a "apt-get install stunnel4" afterwards? Of course you have the same problem without froxlor, but in the latter case, a system admin is responsible for choosing correct usernames. While in the froxlor case, you have a normal user which just picks a username and don't even know it may collide with something like that.

And yes, the correct solution may be that resellers can only create users with a fixed prefix.

d00p:
whooops

d00p:
Hm, after reviewing what has changed, nothing of the descussed changes were added, also i don't get why this is reduced to FCGID but anyway, close

dphi:
@d00p,
* where besides fcgid would this be useful?
* which suggestion is not implemented?

There's the UX question of whether making the prefixes visible in the interface or not, and
how to avoid name collisions  -- both questions are addressed with this patch set.

d00p:
_- where besides fcgid would this be useful?_
Same for php-fpm...also, if you change basic entities like the customer loginname it should be something general, regardless of fcgid/php-fpm/libnss-mysql

_- which suggestion is not implemented?_
The suggestion to make it a "per admin/reseller" setting, see https://github.com/Froxlor/Froxlor/pull/395#issuecomment-259060841

Well, if you set loginnames to be [prefix] + [loginname] the user who uses it should know about this, don't you think?

PR #396, opened 2016-11-13
Feature #1671 Checkbox to enable/disable PHP (vhost settings + fpm po…

schallertd:
Checkbox for enabling/disabling PHP on a per-domain basis.
Reference https://redmine.froxlor.org/issues/1671

Needs some Testing!


install/updates/froxlor/0.9/update_0.9.inc.php
```diff
@@ -3513,3 +3513,14 @@
 	showUpdateStep("Updating from 0.9.38-rc1 to 0.9.38-rc2", false);
 	updateToVersion('0.9.38-rc2');
 }
+
+if (isFroxlorVersion('0.9.38-rc2')) {
```
d00p: do not set new versions...that's what the dbversion is for...just increase that, then set it also in lib/version.inc.php and in install/froxlor.sql (panel_settings)

schallertd:
Btw... the cron job scripts
_./scripts/jobs/cron_tasks.inc.http.10.apache.php
./scripts/jobs/cron_tasks.inc.http.15.apache_fcgid.php
./scripts/jobs/cron_tasks.inc.http.25.lighttpd_fcgid.php
./scripts/jobs/cron_tasks.inc.http.30.nginx.php
./scripts/jobs/cron_tasks.inc.http.35.nginx_phpfpm.php_

already all have checks for 
_if ($domain['phpenabled'] == '1')_

So i guess it had been there already in the past (probably syscp), but had been removed back ten. I simply added the form options and dbmigration to alter `panel_domains` with the old/new field `phpenabled`.

Cheers

d00p:
the fieldname "phpenabled" also comes joined from panel_customers in the webserver-cronjob, see https://github.com/Froxlor/Froxlor/blob/master/lib/classes/webserver/class.WebserverBase.php#L31 either one or the other will be overwritte/ignored

schallertd:
Arghh..  missed that one out, thanks for the hint. Will fix.

d00p:
maybe rename the joined field to something like "phpenabled_global" and then check whether global or domain-specific is 1 (but i'd say do not enable php if global is off, regardless of what the domainsetting is)

schallertd:
Fixed.

schallertd:
Ok, did some testing on two independent froxlor boxes.  
Seems to work fine for me. Maybe some others can confirm.

d00p:
Just a random thought. Sometimes, a customer might want to disable php for a domain himself (like he can enable/disable perl/cgi for example). Should we allow the per-domain php-enabled flag also for customers?

schallertd:
> Just a random thought. Sometimes, a customer might want to disable php for a domain himself (like he can enable/disable perl/cgi for example). Should we allow the per-domain php-enabled flag also for customers?

Well, basically I see no real scenario where a customer himself would like to disable php for one of his domains.  My point of view is that it makes more sense for the admin, especially in combination with the special vhost settings fields  The problem that we had been facing was, that the froxlo's config generation cronjob especially with nginx and php-fpm, always did overwrite much of the custom vhost stuff and made it quite useless. With the new checkbox we are simply able to disable the whole php-fm stuff by froxlor and add our custom vhost config for php, proxying, etc.

I have also thought about a setting where you could completely disable the config cron for a certain domain, but i think this is not a good idea but it's likely to cause more pain with broken vhost configs than it brings.  In my opninion disabling the php-fpm config generation like in this is enough, the rest of the vhost container is anyways just basic stuff like listen block, servername, etc.

Thanks

d00p:
> Btw. please never ever give the customer the possibility to add/edit the additional, custom vhost settings. This would be a headache for every froxlor admin out there... ;-)

We do not intend to :)

Well then, sounds fine to me.

palsch:
Hello guys, 
it looks like we have to enable PHP-FPM to allow vHosts use PHP. If PHP-FPM is desabled, then all vHosts configurations contains following config:

```
  # PHP is disabled for this vHost
  php_flag engine off
```

I enabled PHP checkbox for the hosts.

Please, check that, if the behaviour ist correct here.

Thanks, regards
Paul

d00p:
> it looks like we have to enable PHP-FPM to allow vHosts use PHP.

??? no....

> If PHP-FPM is desabled, then all vHosts configurations contains following config:
> php_flag engine off

Correct when php is disabled for this vhost...

> I enabled PHP checkbox for the hosts.

Please explain more detailed, are you using mod_php or php-fpm now? Please create a thread in our forums (https://forum.froxlor.org) as this is not the right place for support

palsch:
@d00p sorry, i got it now. It was not clear for me, that I have to enable PHP at the customer and at the specific Domain. That would be really nice, if you could add a hint on both sides, that it has to be enabled in two places.

PR #397, opened 2016-11-16
Set FPM-pool-directories per PHP-Config and multiple reload-commands 

Lukas0610:
This would add the possibility to set a custom directory for the FPM-Files for each PHP-Configuration and would allow to use different PHP-Versions with PHP-FPM.

Already tested and confirmed working.


discordier:
Big :+1: on the overall issue, was missing this already. However, I think the path to the fcgi binary needs to be configured in the same way in addition to the fcgi restart command.

Lukas0610:
Committed it right now (in the wrong option at the first try -.-), tested and confirmed working ;)

d00p:
in formfield.phpconfig_add.php and formfield.phpconfig_edit.php it would be way easier if this would be a select-box with the values available from the settings (i would not want to enter a reload command manually every time). Example of how this could be done: https://github.com/Froxlor/Froxlor/blob/master/lib/formfields/admin/domains/formfield.domains_add.php#L50 and https://github.com/Froxlor/Froxlor/blob/master/admin_domains.php#L1035

But very nice indeed. Definetly something for 0.9.39

Lukas0610:
You do not have to enter a restart-command for each template, just for each PHP-FPM. Perhaps letting the user set the path to each FPM-script used and stop/start the FPM's in the code.

d00p:
sorry, i do not seem to get what you mean. Could you please show some screenshots of how the workflow for this feature looks, thanks

Lukas0610:
The setting for the restart-commands looks like this:
As you see, you just need to add the commands to stop all FPM's and then start them again.

d00p:
okaaaaaaay. That's not quite the way froxlor would do this. Why not do it a bit more like FCGID and define the reload-command and config-dir per PHP-configuration?
Then in the cron, when fpm should be restarted, check what configurations are used by domains, grouped by reload-command to get all needed reload-commands.

Also, to avoid the need of first stopping all the daemons, then again start them, we already clean all the configs prior to generation (as you already do in lib/classes/webserver/class.ConfigIO.php).

Lukas0610:
Okay, you would make it this way. To the stop/start: I do it like this because if I change the version from e.g. 7.0 to 5.6, the 5.6-FPM would restart, but the socket would still be in use by the 7.0-FPM and the 5.6 would not start again.

d00p:
True, this would be a problem, but i really don't like having two reload-commands for the same thing...

Lukas0610:
Is this like you thought about it? Already tested and works fine.

discordier:
@d00p How shall we proceed here? I'd love to have this feature to be able to upgrade some customers to PHP 7 and leave some on PHP 5.6 on one box.
So if there is anything we can do to make this PR merge ready just tell us.

scripts/jobs/cron_tasks.php
```diff
@@ -51,6 +52,14 @@
 	 */
 	if ($row['type'] == '1') {
 
+		// get class for phpfpm-restart
+		$phpfpm_restart = new phpfpm_restart($cronlog);
+
+		// if phpfpm is enabled, capture current php-configs
+		if (Settings::Get('phpfpm.enabled') == 1) {
+			$phpfpm_restart->captureCurrentConfigs();
```
d00p: why is that needed?

templates/Sparkle/admin/phpconfig/overview.tpl
```diff
@@ -20,6 +20,7 @@ $header
 						<th>{$lng['admin']['phpsettings']['description']}</th>
 						<th>{$lng['admin']['phpsettings']['activedomains']}</th>
 						<th>{$lng['admin']['phpsettings']['binary']}</th>
+						<th>{$lng['admin']['phpsettings']['configdir']['title']}</th>
```
d00p: why display config-directory but not the runscript?

scripts/jobs/cron_tasks.php
Lukas0610: This is required if e.g. multiple configs use the same restart-script and to restart only servers which are used and leave unused ones untouched. It seems weird because it was planned otherwise, I will change some things

lng/english.lng.php
```diff
@@ -2062,6 +2062,9 @@
...
-
 $lng['serversettings']['nginx_http2_support']['title'] = 'Nginx HTTP2 Support';
 $lng['serversettings']['nginx_http2_support']['description'] = 'enable http2 support for ssl. ENABLE ONLY IF YOUR Nginx SUPPORT THIS FEATURE. (version 1.9.5+)';
+$lng['admin']['phpsettings']['configdir']['title'] = 'Configuration directory';
+$lng['admin']['phpsettings']['configdir']['description'] = 'Directory which contains the php-fpm configurations. If empty the default directory will be selected';
+$lng['admin']['phpsettings']['runscript']['title'] = 'FPM run-script-path';
+$lng['admin']['phpsettings']['runscript']['description'] = '<strong>Without reload, restart, stop or start</strong>, just the path.<br/> Script which will be used to restart the FPM. If empty, the default command will be executed';
```
Lukas0610: Jep, just **/etc/init.d/php7-fpm** would be right. Because it appends start/stop automatically in the cronjob

d00p: So again, the description "just the PLATH" is wrong...

d00p: *Path

Lukas0610: How would you title it then? It's not the whole command but just like the description says the path of the script which is usually used to start/stop the fpm-server

d00p: "Specify the init-script without parameters, e.g. '**/etc/init.d/php5-fpm**'"

chaos234:
Any responses to this feature @d00p when it will be live in froxlor?

@Lukas0610 as long as this isn't officially implemented, would it be safe to implement it on my own (and on my own risk)? Would be nice to have this feature. Also tell me please if I must change things on my server settings (debian, php5-fpm, nginx and so far).

d00p:
Ich would Not use this. Ich have a better Idea for doing this. But simply No Time

chaos234:
Sounds bad but if your time is rare than nothing can be done to this :/

d00p:
implemented otherwise, see https://github.com/Froxlor/Froxlor/commit/af55fe5b82593d4fba6f71ed9ad062ec5431c453 and following

PR #398, opened 2016-11-17
Add support for http2 option to nginx

hypernics:
I tested this modification and works great. 
I added the setting manually to my panel_settings table. I know nothing how to make the database update in update_0.9.inc.php file, so it commented out the code what I believe to correct.


install/updates/froxlor/0.9/update_0.9.inc.php
```diff
@@ -3513,3 +3513,13 @@
 	showUpdateStep("Updating from 0.9.38-rc1 to 0.9.38-rc2", false);
 	updateToVersion('0.9.38-rc2');
 }
+/*
+if (isDatabaseVersion('201610070')) {
```
d00p: Update looks correct, but when "active" use the current date and not 201610270 as new dbversion :)

d00p:
Apart from the update this does look good to me

PR #400, opened 2016-11-25
Better handling for letsencrypt errors

micw:
Better handling for letsencrypt errors after failed registration or changed license:

* try to re-register with last known key if registration failed before
* try to re-accept terms if it failed before or if terms are changed
* trigger re-registration with last known key if certificate generation fails with 403 Unauthorized


lib/classes/ssl/class.lescript.php
```diff
@@ -537,10 +591,46 @@ public function getLastLinks()
...
+		$license=$this->getAgreementURLFromDirectory();
+		if (!empty($license)) return $license;
+
+		// 3. query /terms endpoint (not ACME standard but implemented by let's enrypt)
+		$license=$this->getAgreementURLFromTermsUrl();
+		if (!empty($license)) return $license;
+
+		// Fallback: use latest known license. This is only valid for let's encrypt and should be removed as soon as there is an official
+		// ACME-endpoint to get the current ToS
+		return "xxxhttps://letsencrypt.org/documents/LE-SA-v1.1.1-August-1-2016.pdf]";
```
d00p: "xxx" at the beginning? "]" at the end? This would not work

micw: Oops, sorry that was left from my error handling tests. The rest is tested, yes.

PR #404, opened 2016-12-17
Fix website not loading in current Firefox ESR 45 and tweak webserver config file according to Mozillas recommendations

pissbeutel:
See: https://wiki.mozilla.org/Security/Server_Side_TLS

With the old cipher suite (ECDH+AESGCM:ECDH+AES256:!aNULL:!MD5:!DSS:!DH:!AES128), Firefox ESR 45 will stop loading the page and the window will stay blank (tested with nginx).
The config files for apache and nginx were updated as per Mozilla's recommendation, the lighttpd config is still good.
This will improve the SSLLabs score as well.

### DIFF
--- a/actions/admin/settings/131.ssl.php
+++ b/actions/admin/settings/131.ssl.php
@@ -36,7 +36,7 @@ return array(
 					'varname' => 'ssl_cipher_list',
 					'type' => 'string',
 					'string_emptyallowed' => false,
-					'default' => 'ECDH+AESGCM:ECDH+AES256:!aNULL:!MD5:!DSS:!DH:!AES128',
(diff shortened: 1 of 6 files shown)


schallertd:
This isn't as easy as it might sound, because the available cipher suites depends on the installed openssl version.  But ofc. you can easily change this yourself within the SSL-Settings of Froxlor. In general i'd always suggest to use very moderate cipher suites as Froxlor's defaults, as this could may lead to issues when the openssl version is too old.

For example we're using the following cipher suites with OpenSSL 1.0.1t, which provide us with a good balance between security and supporting older clients;

ssl_ciphers `ECDHE-RSA-AES128-GCM-SHA256:ECDHE-RSA-AES256-GCM-SHA384:ECDHE-RSA-AES128-SHA:ECDHE-RSA-AES256-SHA:ECDHE-RSA-AES128-SHA256:ECDHE-RSA-AES256-SHA384:DHE-RSA-AES128-GCM-SHA256:DHE-RSA-AES256-GCM-SHA384:DHE-RSA-AES128-SHA:DHE-RSA-AES256-SHA:DHE-RSA-AES128-SHA256:DHE-RSA-AES256-SHA256:EDH-RSA-DES-CBC3-SHA:DES-CBC3-SHA`;

d00p:
No more action from op for 2 months after change request -> close

PR #406, opened 2017-01-05
Turkish language added

emresaracoglu:
Turkish lang file added for Froxlor
I will start translating to Turkish language


d00p:
why did you create this pull request then if you haven't even started yet? Please put only turkish strings in that file, everything that is missing is used from the english language file as fallback (so no need to have the english strings in your turkish language file!)

Re-open when done, not when starting, thanks

PR #408, opened 2017-01-17
WIP: SSL OCSP Stapling

buffcode:
"Allow client software using SSL to communicate with the server to efficiently check that the server certificate has not been revoked."

Source:
https://wiki.apache.org/httpd/OCSPStapling

- [x] Configuration settings (defaults to off, because cache path may be system dependent)
- [x] Output in vhosts configuration
- [ ] Installation / update scripts
  - [x] Database fields for installation
  - [ ] Update script
- [ ] Translations
  - [ ] Dutch
  - [x] English
  - [ ] French   
  - [x] German
  - [ ] Italian
  - [ ] Portugues
  - [ ] Swedish

### DIFF
--- a/actions/admin/settings/131.ssl.php
+++ b/actions/admin/settings/131.ssl.php
@@ -153,6 +153,23 @@ return array(
 					'type' => 'bool',
 					'default' => false,
 					'save_method' => 'storeSettingField'
+				),
+				'system_ssl_use_stapling' => array(
+					'label' => $lng['serversettings']['ssl']['use_stapling'],
+					'settinggroup' => 'system',
+					'varname' => 'ssl_use_stapling',
+					'type' => 'bool',
+					'default' => true,
+					'save_method' => 'storeSettingField'
+				),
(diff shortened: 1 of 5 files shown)


buffcode:
How are updates to the `update_0.9.inc.php` handled? Should they be included in the commit or will some maintainer update those scripts with the correct versions?

Do all translations need to be provided for merge? Google Translate?

d00p:
- [ ] nginx support
- [ ] lighttpd support
- [ ] apache completety, you only added the directives for the ip/port vhosts, not for customers
- [ ] please do not use **#** as inline comment-starter, it's **//**

d00p:
> How are updates to the update_0.9.inc.php handled?

Take a look at the file

> Should they be included in the commit or will some maintainer update those scripts with the correct versions?

You never change the version, you change the db_version, and yes, you need to include updates

> Do all translations need to be provided for merge?

No

> Google Translate?

No, never!

PR #409, opened 2017-01-17
RFC: Code styling

buffcode:
Code style configuration via [PHP Coding Standards Fixer](https://github.com/FriendsOfPHP/PHP-CS-Fixer).
I tried to extract the currently used style from the project files and set it up using PHP CS Fixer.

 Installation:
 - `composer install`

Notes:
- Changes to existing files will occur, as the code style is very heterogeneous and contains indentation errors etc.
- To apply/fix the style: `php-cs-fixer fix --config=.php_cs.dist --verbose "<fileOrPath>"`

Please comment if the current settings are fine. More aggressive/restrictive settings would be desirable but would change lots of code. The settings should be upgraded in the future to apply more fixes (such as removal/replacement of deprecated functions/APIs, migrations for PHP7 etc.)

Recommended: Automated TravisCI testing for code style issues

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,8 @@ logs/*
 .well-known
 .idea
 *.iml
+.php_cs.cache
+/vendor
 
 !templates/Froxlor/
 !templates/Sparkle/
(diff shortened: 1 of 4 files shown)


d00p:
Why don't you TALK to us first maybe? We're all on IRC. Also, you didn't even explain WHY you are doing this? Just for a bit of code-formatting?

PR #410, opened 2017-01-18
0.99

Toppi68:
Hi guys,
i noticed, that i can not add/change folder-protections anymore.

See here in: customer_extras.php Line: 123
$authname = validate($_POST['directory_authname'], 'directory_authname', '/^[a-zA-Z0-9][a-zA-Z0-9\-_ ]+\$?$/');

This regEx makes no sense imo - in example i am using "example.com" or "domain.de" as basedirs.
Its not working anymore because the \. is missing

Thank you very much guys.
Toppi


d00p:
Wrong branch mate...

PR #411, opened 2017-01-28
Admin database manager, list and create database

hypernics:
Admin module to list databases and create database to your customers.
It allow to create database with custom name (as I usually do) or use the generated name.


d00p:
a complete new module with lots of code already existing for something you can do by just switching to the customer...except the custom-naming which is not really intended...

hypernics:
It's your project, your choice. The code is almost copy-pasted from customer mysql manager code and admin domain manager code.
The custom naming is very useful for me.
If you merge it i am happy to help the community. If you don't want it then close the pull request and I will merge into my own fork.

lib/classes/database/class.DbManager.php
```diff
@@ -1,127 +1,151 @@
-<?php
```
discordier: What are these changes here? Did you by accident change the line endings?

PR #413, opened 2017-02-01
Internal IP and port

HitkoDev:
Add internal IP/Port config option, useful when the public server IP (used by nameserver) is different than the webserver IP (e.g. ```<VirtualHost 0.0.0.0:80>```). This is useful when using AWS Elastic IP or DO floating IP, or when the server is behind a proxy.


d00p:
You should at least add an enable/disable setting for that so these fields can be hidden if not activated. It's irritating for people who do not need this (e.g. most of the users)

d00p:
Nice idea basically but this PR just replaces the ipport values with its internalipport values making this a total screwup when using....

PR #415, opened 2017-03-22
Disable auto-generating A or AAAA record with same name if user set one

Alsik:
On dual-stack hosting Froxlor auto-generate two (or more) required DNS records and there is no possibility to set only one custom record and remove second automatic record. This patch disable auto-generating A or AAAA record with same name if user set one.


d00p:
And what about people that just want to overwrite one of the two values but keep the other?

Alsik:
Then must set another one manually. Is more common point one name to IPv4 and IPv6 adresses on same host than split into more. And still is more common have IPv4 address than IPv6. With Froxlor on dual-stack you cannot point name to only IPv4 services.

d00p:
I just don't really like that hidden behaviour - i mean, required A/AAAA records are what the customer added in his domainssettings (e.g. added subdomains etc.) - so overwriting the A/AAAA record makes no real sense - better remove the subdomain then. Or in case of A <> AAAA - just remove the ipv6 address from the domainsettings as admin.

Alsik:
I see. Another case is MX record, but admin can remove emaildomain in settings. And customers with enabled DNS editor must take care if edit A/AAAA autogenerated record. Thanks for clarify this.

PR #418, opened 2017-04-03
Correct Wiki links #2

ianklemm:
The changes regarding the "plaintext auth" can be ignored, my bad, I forgot the create a new branch for the wiki link fix.


d00p:
Please do not mix the correction of the Wiki links with changes in the config templates

PR #422, opened 2017-04-10
add OCSP stapling support for apache2 and nginx

nachtgeist:


nachtgeist:
good catch! updated my master.

PR #424, opened 2017-04-12
Removed query parameter s, used cookies instead

shyim:
I guess everyone hates the sessionid in the query. Best example is go back to a specifc site with changing the request uri, if i forget the session i will be logged out or bookmarking specifc sites its impossible :(


d00p:
Already done for 2.0, i dont really want such huge changes in 0.9.x - other thoughts @Dessa @nachtgeist @BNoiZe ?

nachtgeist:
what d00p said...

shyim:
And why is the 2.0 branch not available on github.................

d00p:
Because of various reasons. It will be when it reaches the wanted state.

PR #427, opened 2017-04-14
mysql 5.7, Remove unsupported default value in text column

sbayd:
For fixing installation on mysql 5.7 & php 7, I have to remove default value from text field otherwise table is not getting created.
https://dev.mysql.com/doc/refman/5.7/en/blob.html
BLOB and TEXT columns cannot have DEFAULT values.


d00p:
The problem is STRICT Mode, not only the default

d00p:
refs ticket #433

d00p:
See 1e03946df7e0a8aebe8d2c680b9349be81dafda3

PR #429, opened 2017-04-15
Skip views when altering table charsets

mkhon:
"ALTER TABLE ... CONVERT TO CHARACTER SET" fails on VIEWs with error like

ERROR 1347 (HY000): 'froxlor.view_xxx' is not BASE TABLE

Views can be created for third-party extensions and API to Froxlor database (see for example opendkim support).


d00p:
Neat ;)

PR #430, opened 2017-04-17
Fix apache24_ocsp_cache_path definition

mkhon:
Add default value for apache24_ocsp_cache_path, otherwise saving
SSL settings for !apache24 fails on apache24_ocsp_cache_path validation.


mkhon:
screen shot 2017-04-17 at 15 10 34

PR #443, opened 2017-05-29
Set permissions of generated files

RipClaw2971:
Set permissions of the new generated files to

passwd 0644 root root
group 0644 root root
shadow 0600 root root

The permissions of the shadow file is for security. The other ones are just to get the right permissions and avoid problems with umask on a system

### DIFF
--- a/scripts/classes/class.Extrausers.php
+++ b/scripts/classes/class.Extrausers.php
@@ -53,12 +53,21 @@ class Extrausers
 			switch ($type) {
 				case 'passwd':
 					$line = $u['username'] . ':' . $u['password'] . ':' . $u['uid'] . ':' . $u['gid'] . ':' . $u['comment'] . ':' . $u['homedir'] . ':' . $u['shell'] . PHP_EOL;
+					$permissions = 0644;
+					$user = 'root';
+					$group = 'root';
 					break;
 				case 'group':
 					$line = $u['groupname'] . ':' . $u['password'] . ':' . $u['gid'] . ':' . $u['members'] . PHP_EOL;
+					$permissions = 0644;
+					$user = 'root';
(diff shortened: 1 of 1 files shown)


scripts/classes/class.Extrausers.php
```diff
@@ -68,5 +77,20 @@ private static function _generateFile($file, $query, &$cronlog)
 		} else {
 			$cronlog->logAction(CRON_ACTION, LOG_NOTICE, 'Error when writing ' . $type . ' file entries');
 		}
+		if (chmod($file,$permissions) !== false) {
```
d00p: I don't t hink you need to log when the permission-setting is successful (we almost never do).

scripts/classes/class.Extrausers.php
```diff
@@ -53,12 +53,21 @@ private static function _generateFile($file, $query, &$cronlog)
 			switch ($type) {
 				case 'passwd':
 					$line = $u['username'] . ':' . $u['password'] . ':' . $u['uid'] . ':' . $u['gid'] . ':' . $u['comment'] . ':' . $u['homedir'] . ':' . $u['shell'] . PHP_EOL;
+					$permissions = 0644;
+					$user = 'root';
```
d00p: user is always root, also hardcoded ...as this script is run as cron with user root, the owner and group of newly created files should be the root user anyway...superfluous imho

nachtgeist: Doesn't hurt, either. Since you're not unlinking it if it exists but only touching it if it does not, it's possible that it pre-exists belonging to something != root:root...

d00p: agreed

PR #450, opened 2017-06-10
nginx: add index directive to path options for secured directories

### LINKED ISSUE #449: nginx: index directive is not applied when setting directory security
# Summary

When setting directory security via Froxlor, the `index` directive is not applied. Thus, accessing a path does not automatically load `index.php` (or `index.htm`) if a filename is omitted.

# System information
* Froxlor version: 0.9.38.7
* Web server: nginx

# Proposed solution

I want to supply a patch for this issue, but wanted to get opinions on the preferred solution first:

1. add a new `index` directive for every `location` block created for directory security
2. add `location` blocks under the main `location /` block to enable inheritance for location settings

Solution 1 would mean minimal changes while 2 would automatically solve issues that arise from similar circumstances.

oliverrahner:
`index` directives are now created for secured directory locations, closes #449


d00p:
thank you very much. Patch tested?

oliverrahner:
Yes, tested by adding this patch to my own production installation and testing my various vHost.

PR #451, opened 2017-06-10
add setting for webserver log format for nginx and Apache

oliverrahner:
adding an admin definable setting for webserver log format to replace the default 'combined' format for both nginx and Apache

Patch has been tested for the case of upgrade installation, no new installation has been tested!


d00p:
What about lighttpd? Also mit would be helpful to Show the default value and possibly a link to the Apache doc so people know what values are possible.

A possible problem: webalizer/AWStats won't know about the customers Format, hence the Traffic calculation could be wrong / Not working.

PR #459, opened 2017-07-07
switch file_get_contents with php curl

pquerner:
Attempt to improve lescript workflow by using PHP's curl, instead of PHP's file_get_content.

### DIFF
--- a/lib/classes/ssl/class.lescript.php
+++ b/lib/classes/ssl/class.lescript.php
@@ -227,10 +227,9 @@ class lescript
 			$this->log("Token for $domain saved at $tokenPath and should be available at $uri");
 
 			// simple self check
-			$selfcheckContextOptions = array('http' => array('header' => "User-Agent: Froxlor/".$this->version));
-			$selfcheckContext = stream_context_create($selfcheckContextOptions);
-			if ($payload !== trim(@file_get_contents($uri, false, $selfcheckContext))) {
-				$errmsg = json_encode(error_get_last());
+			$payload = $this->getUrlContent($uri);
+			if (isset($payload['error']) && $payload['error'] === TRUE) {
+				$errmsg = isset($payload['error_message']) ? $payload['error_message'] : json_encode(error_get_last());
 				if ($errmsg != "null") {
 					$errmsg = "; PHP error: " . $errmsg;
 				} else {
@@ -246,7 +245,7 @@ class lescript
 			$result = $this->signedRequest($challenge['uri'], array(
 				"resource" => "challenge",
 				"type" => "http-01",
-				"keyAuthorization" => $payload,
+				"keyAuthorization" => isset($payload['data']) ? $payload['data'] : '',
 				"token" => $challenge['token']
 			));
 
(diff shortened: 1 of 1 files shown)


lib/classes/ssl/class.lescript.php
```diff
@@ -494,6 +493,35 @@ protected function log($message)
...
+     */
+    protected function getUrlContent($url){
+        $r = [
+            'error' => TRUE,
+            'error_message' => '',
+            'data' => '',
+        ];
+        $ch = curl_init();
+        curl_setopt($ch, CURLOPT_URL, $url);
+        curl_setopt($ch, CURLOPT_USERAGENT, 'Mozilla/4.0 (compatible; MSIE 6.0; Windows NT 5.1; .NET CLR 1.1.4322)');
```
d00p: why not use the same as in update-check?

d00p:
tested?

pquerner:
Yes, I tested it on 2 domains.

lib/classes/ssl/class.lescript.php
pquerner: Are talking about the Useragent?

lib/classes/ssl/class.lescript.php
```diff
@@ -494,6 +493,35 @@ protected function log($message)
...
+        curl_setopt($ch, CURLOPT_RETURNTRANSFER, 1);
+        $data = curl_exec($ch);
+        $httpcode = curl_getinfo($ch, CURLINFO_HTTP_CODE);
+        if($httpcode>=200 && $httpcode<300) {
```
pquerner: This could be changed to a preg match, if we know how the challenges must look upfront. (I'm sure there is some specc around from LE)

lib/classes/ssl/class.lescript.php
d00p: You even patched it away, it was in there already: https://github.com/Froxlor/Froxlor/blob/master/lib/classes/ssl/class.lescript.php#L230

lib/classes/ssl/class.lescript.php
d00p: it's a simple SELF-check whether the acme-alias works or not...has nothing to do with Let's Encrypt and it's specs

lib/classes/ssl/class.lescript.php
pquerner: Yea, but I mean what does it really matter? Do you check the logs from calls you do yourself? I mean, if you really want it in there I'll change it back.

lib/classes/ssl/class.lescript.php
pquerner: You use the acme response later on, no? So you _could_ check it against some spec.

lib/classes/ssl/class.lescript.php
d00p: it doesnt really matter, but i don't like a windows user-agent ...when hardcoding something that doesnt matter, let it be something project pecific not something that could be real

lib/classes/ssl/class.lescript.php
d00p: later on, yes...but we won't even bother sending the request to let's encrypt if the self-check fails, preventing ratelimits if misconfigured...please do not change working features. Enhance, gladly, but please understand what you do

pquerner: https://github.com/Froxlor/Froxlor/blob/master/lib/classes/ssl/class.lescript.php#L246
You do send out that challenge right away, when you could check if that even makes sense. 
$result can have something like "too many requests by IP", so why would you annoy LE if you dont have correct data from your acme-challenge to begin with.

lib/classes/ssl/class.lescript.php
pquerner: Ok, I can understand that. Will change it asap.

pquerner: Changed in https://github.com/Froxlor/Froxlor/pull/459/commits/603b959805961bcd93c4f5e564a785174c149371

lib/classes/ssl/class.lescript.php
pquerner: Spec should be this one, if @d00p acknowledges that work should continue on this behalf. 
https://github.com/ietf-wg-acme/acme/blob/master/draft-ietf-acme-acme.md#http-challenge

d00p: you replaced the SELF-CHECK...there just is NO need to do anything let's encrypt related...see https://github.com/Froxlor/Froxlor/pull/459#discussion_r126204320

pquerner: OK, fine with that, since #460 is merged.

discordier:
I'd prefer a failover solution where a check is added if `curl_exec()` succeeds and if not, revert to `file_get_contents()`.

Would be bc for installations where curl is either not available or disabled.

d00p:
Curl is required anyway for Let's encrypt

pquerner:
According to docs php-curl is a dependency.

d00p:
Don't forget that this patch is ONLY about the selfcheck, nothing with communicating with LE at all

discordier:
Ah, ok. Then scrap my comment. :smile:

d00p:
there's no real reason to change this

PR #494, opened 2017-12-08
extend nginx letsencrypt (.well-known) redirect regex to https

xelaz:
I extend the regex rule for .well-known location to redirect to https site, because this one was not enough:
if ($request_uri !~ ^/.well-known/acme-challenge/\w+$)

Some Keys from LetsEncrypt cold not redirected because there are "-" signs in Key
Key Example: WgV7aLbGyXrdnFHCIIssDE_914eI9EYwRAoFc5Ci3-w

New check is this:
if ($request_uri !~ ^/.well-known/acme-challenge/**[\w-]**+$) {

### DIFF
--- a/scripts/jobs/cron_tasks.inc.http.30.nginx.php
+++ b/scripts/jobs/cron_tasks.inc.http.30.nginx.php
@@ -224,7 +224,7 @@ class nginx extends HttpConfigBase
 					} else {
 						$_sslport = $this->checkAlternativeSslPort();
 						$mypath = 'https://' . Settings::Get('system.hostname') . $_sslport . '/';
-						$this->nginx_data[$vhost_filename] .= "\t" . 'if ($request_uri !~ ^/.well-known/acme-challenge/\w+$) {' . "\n";
+						$this->nginx_data[$vhost_filename] .= "\t" . 'if ($request_uri !~ ^/.well-known/acme-challenge/[\w-]+$) {' . "\n";
 						$this->nginx_data[$vhost_filename] .= "\t\t" . 'return 301 ' . $mypath . '$request_uri;' . "\n";
 						$this->nginx_data[$vhost_filename] .= "\t" . '}' . "\n";
 					}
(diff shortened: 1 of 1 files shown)


d00p:
tested?

scripts/jobs/cron_tasks.inc.http.30.nginx.php
```diff
@@ -224,7 +224,7 @@ public function createIpPort()
 					} else {
 						$_sslport = $this->checkAlternativeSslPort();
 						$mypath = 'https://' . Settings::Get('system.hostname') . $_sslport . '/';
-						$this->nginx_data[$vhost_filename] .= "\t" . 'if ($request_uri !~ ^/.well-known/acme-challenge/\w+$) {' . "\n";
+						$this->nginx_data[$vhost_filename] .= "\t" . 'if ($request_uri !~ ^/.well-known/acme-challenge/[\w-]+$) {' . "\n";
```
nachtgeist: To avoid mistakes/ambiguities on further extensions of the regex, with exception to the negation character ^ a dash should *always* be the first character within a character class.
--> `[-\w]+`

scripts/jobs/cron_tasks.inc.http.30.nginx.php
```diff
@@ -473,7 +473,7 @@ protected function getVhostContent($domain, $ssl_vhost = false)
 			// Get domain's redirect code
 			$code = getDomainRedirectCode($domain['id'], '301');
 
-			$vhost_content .= "\t" . 'if ($request_uri !~ ^/.well-known/acme-challenge/\w+$) {' . "\n";
+			$vhost_content .= "\t" . 'if ($request_uri !~ ^/.well-known/acme-challenge/[\w-]+$) {' . "\n";
```
nachtgeist: idem

xelaz:
yes it works!

xelaz:
An i have another problem if i use only nginx and letsencrypt.

If you enable letsencrypt then you can set your config path to the acme.conf and this is very tricky. 
At first i destroy my configs while the default value is:
_/etc/apache2/config-enabled/acme.conf_
After i found the Problem i paste my own config to nginx:
_/etc/nginx/snippets/acme.conf_

```
location ~ ^/\.well-known/acme-challenge/([-\w]+)$ {
    default_type "text/plain";
    allow all;
    alias /var/www/letsencrypt/.well-known/acme-challenge/$1;
    break;
}
```
This is my dirty solution to break the redirect if you choice only https in your domain.

nachtgeist:
Sorry @d00p, miscommunication on my side.

Anyways, with #561 reverted, you can merge this one instead.

Stopi:
I recently had a failure with this.
It was working for a bit but then fails suddenly for one hostname.

After debugging, I finally solved the issue by adding a **colon** to the regex.
`^/\.well-known/acme-challenge/([-:\w]+)$`
Certbot was trying to access a token ending with a colon.

I suppose it doesn't matter but it was failing with one of my apache server, I still haven't got this issue on nginx.

nachtgeist:
Hm. How about we change the regex to ```^/\.well-known/acme-challenge/([^/]+)$``` ?

PR #506, opened 2018-01-18
http2 could be set even if http2 is not enabled

RipClaw2971:
http2 could be set if you add or edit a domain even if http2 is disabled in the settings

### DIFF
--- a/lib/formfields/admin/domains/formfield.domains_add.php
+++ b/lib/formfields/admin/domains/formfield.domains_add.php
@@ -174,7 +174,7 @@ return array(
 						'value' => array()
 					),
 					'http2' => array(
-						'visible' => ($ssl_ipsandports != '' ? true : false) && Settings::Get('system.webserver') != 'lighttpd',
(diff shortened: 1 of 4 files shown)


lib/formfields/admin/domains/formfield.domains_add.php
```diff
@@ -174,7 +174,7 @@
 						'value' => array()
 					),
 					'http2' => array(
-						'visible' => ($ssl_ipsandports != '' ? true : false) && Settings::Get('system.webserver') != 'lighttpd',
+						'visible' => ($ssl_ipsandports != '' ? true : false) && Settings::Get('system.webserver') != 'lighttpd' && Settings::Get('system.http2_support') == '1',
```
d00p: no need to check for webserver != 'lighttpd' imho because the setting is not available for lighttpd and thus cannot be activated anyway. Also, in case lighttp includes http2 support this would be a hard-to-find-thing. Just a check for http2-support should be enough

d00p:
Just the tiny correction with webserver != 'lighttpd' and this is good to go, thanks

PR #523, opened 2018-02-21
0.10.0 - API get(), delete() with username

JB1985:
Let the API have access to use get(), delete() with username.


d00p:
make thiis an issue please, not a PR requesting to merge 0.10.0 to master

PR #525, opened 2018-02-24
Added creation of fullchain certs based on letsencrypt certs, i.e. fo…

tnebe:
…r usage in postfix

### DIFF
--- a/lib/classes/webserver/class.DomainSSL.php
+++ b/lib/classes/webserver/class.DomainSSL.php
@@ -77,6 +77,7 @@ class DomainSSL {
 			// initialize optional files
 			$ssl_files['ssl_ca_file'] = '';
 			$ssl_files['ssl_cert_chainfile'] = '';
+			$ssl_files['ssl_cert_fullchainfile'] = '';
 			// set them if they are != empty
 			if ($dom_certs['ssl_ca_file'] != '') {
 				$ssl_files['ssl_ca_file'] = makeCorrectFile($sslcertpath.'/'.$domain['domain'].'_CA.pem');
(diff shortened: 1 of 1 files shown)


lib/classes/webserver/class.DomainSSL.php
```diff
@@ -89,6 +90,10 @@ public function setDomainSSLFilesArray(array &$domain = null) {
 					$ssl_files['ssl_cert_chainfile'] = makeCorrectFile($sslcertpath.'/'.$domain['domain'].'_chain.pem');
 				}
 			}
+            		if (($dom_certs['ssl_ca_file'] != '') and ($dom_certs['ssl_cert_chainfile'] != '')) {
```
d00p: and what if only one of these is set? What if another services config relies on the file **$sslcertpath.'/'.$domain['domain'].'_fullchain.pem'** to exist? The fullchain does not necessarily require both, ca and chainfile to be filled....with one of them, it's still the fullchain

tnebe: OK, i´m not that save with specification of fullchain certs, but can easily change the condition to `or` and check only existing one of them.
Apart from combining them manually: Isn´t fullchain not already retrieved from LE already but not stored in a file?
https://github.com/Froxlor/Froxlor/blob/4d3fa6eca510329542ec4518c4bc029533051d01/lib/classes/ssl/class.lescript.php#L331

d00p:
done in https://github.com/Froxlor/Froxlor/commit/2b366c8f23317ccab87f873af999147e7d05dd5e

PR #528, opened 2018-03-02
Fixed SQL Error due missing fullchain

shyim:
Hey,
the last commit has introduced a sql error when requesting a letsencrypt cert. This pr adds the missing fullchain parameter.

```
froxlor[602]: SQLSTATE[HY093]: Invalid parameter number: number of bound variables does not match number of tokens
froxlor[602]: --- DEBUG: #0 /var/www/lib/classes/database/class.Database.php(72): PDOStatement->execute(Array) #1 /var/www/scripts/jobs/cron_letsencrypt.php(263): Database::pexecute(Object(PDOStatement), Array) #2 /var/www/scripts/froxlor_master_cronjob.php(80): require_once('/var/www/script...') #3 {main}
```

### DIFF
--- a/scripts/jobs/cron_letsencrypt.php
+++ b/scripts/jobs/cron_letsencrypt.php
@@ -260,6 +260,7 @@ foreach ($certrows as $certrow) {
 				'ca' => $return['chain'],
 				'chain' => $return['chain'],
 				'csr' => $return['csr'],
+				'fullchain' => $return['fullchain'],
 				'expirationdate' => date('Y-m-d H:i:s', $newcert['validTo_time_t'])
 			));
 


d00p:
You are right, same with v2

Crease29:
Is there a reason why you don't merge it, @d00p ? I mean you just pushed the same fix on your own. :unamused:

d00p:
because waiting for you to update the PR takes longer then just pushing and closing in the commit message...no offense man

teiling88:
very community friendly...

d00p:
> very community friendly...

@teiling88 if you were the one waiting for the feature, you would be the first to moan why i don't merge this...history shows my decision is right on things like that because everytime I give feedback to a pull-request nothing happens for days or even weeks...

So, if you don't like it, fine, I can live with that but don't you "very community friendly" me - I give more support in the forum than anything; Sure I'm not nice - but I want people to learn from their mistakes and think for themselves - doing everything for them does not help with that.

Why is it that people always cry about bad things but noone ever just says "thanks"...think about that man, I don't get paid for this...

PR #529, opened 2018-03-03
Froxlor installer: Use same file path for chmod.

michaelzangl:
Just noticed this wile automating installation.


d00p:
Thanks, very nice.

PR #530, opened 2018-03-03
Fix hostname error message

michaelzangl:
Hostname is validated to be a valid domain name.

I just fixed the error message. I don't know if strict validation is that useful - it makes it impossible to install froxlor on localhost / any other local TLD without tricks.


d00p:
It's intended that a FQDN is specified...thanks for the clearer error-strings

PR #531, opened 2018-03-04
Link domains to HTTPS in certificate list

Crease29:
As an admin I want to check if the certificate of a website is working.
Therefore I click on the domain in the SSL certificate list and have to prepend "https" to the URL later.
In this case I'd link the domain to HTTPS initially.


d00p:
Good point. Thanks :)

PR #541, opened 2018-03-29
PHP Codestyle Fix

fionera:
Running php-cs-fixer fixes codestyle issues and updates it to the 21th century


d00p:
a bit more information please...?! Formatting code does not put it in the 21th century :P You may want to check the https://github.com/Froxlor/Froxlor/tree/0.10.0 branch and put the effort into the next version instead of 0.9.x

fionera:
I just ran PHP-CS because the Code looks horrible at some places. Please fix the Codestyle and move to min version php5.6

Dessa:
moving to 5.6 is not possible at the moment., centos still has 5.4 as default.

d00p:
also, why move to php-5.6 if there's no programmical reason for it? There are no constructs, commands or whatever we need right now that require min. 5.6 - Plus I don't understand what the php-min version has to do with codestyle...
As said before, you might want to put your effort of codestyling etc. to our upcoming version 0.10.0

d00p:
no more answers

PR #566, opened 2018-08-15
Fix dovecot and postfix getting incorrect information about an user with IDN domain

continga:
With dovecot's password_query selecting username as email, we were overwriting the puny-encoded domain a client delivered with the real unicode one (see https://wiki2.dovecot.org/PasswordDatabase/ExtraFields ), which currently doesn't create problems (because e.g. we are selecting the `home` directory independently of the username and take it from the DB), but it is imho incorrect nontheless.

In postfix's virtual_sender_permission however, this was creating problems when the reject_sender_login_mismatch sender restriction was being used, as the login-name of the user (punycode) was not equal to the login-name (unicode) that is allowed to send with that email-address.

Note: In my usecases with Froxlor, the `username` and `email` fields in the `mail_users` table always contained exactly the same values, except for IDN domains - so I am a bit confused, why we even have these 2 different fields... I currently cannot think of other use-cases that produces different values in the `username` and `email` fields, so the changes here made by me should be safe to merge. BUT: It would be nice if someone more experienced with Froxlor could tell whether this maybe would break other setups, where the `username` and `email` fields legitimately were different, and if yes how we then could fix this IDN-problem here

### DIFF
--- a/lib/configfiles/gentoo.xml
+++ b/lib/configfiles/gentoo.xml
@@ -1536,7 +1536,7 @@ user = <SQL_UNPRIVILEGED_USER>
 password = <SQL_UNPRIVILEGED_PASSWORD>
 dbname = <SQL_DB>
 hosts = <SQL_HOST>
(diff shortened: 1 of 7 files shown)


d00p:
basically these are separated so the username COULD be something other then the email address

PR #572, opened 2018-09-17
Fixed class.lescript_v2.php to work with ACMEv2

markuspetermann:
When requesting a certificate for multiple Domains/SANs, the request must be done in a single order.

### DIFF
--- a/install/updates/froxlor/0.9/update_0.9.inc.php
+++ b/install/updates/froxlor/0.9/update_0.9.inc.php
@@ -3998,3 +3998,16 @@ if (isDatabaseVersion('201805241')) {
 		updateToDbVersion('201805290');
 	}
 }
+
+if (isDatabaseVersion('201805290')) {
+
+	showUpdateStep("Adding leaccount field to panel customers");
+	Database::query("ALTER TABLE `" . TABLE_PANEL_CUSTOMERS . "` ADD COLUMN `leaccount` varchar(255) default '' AFTER `leregistered`;");
+	lastStepStatus(0);
+
+	showUpdateStep("Adding system setting for let's-encrypt account");
+	Settings::AddNew('system.leaccount', "");
+	lastStepStatus(0);
+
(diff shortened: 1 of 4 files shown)


install/updates/froxlor/0.9/update_0.9.inc.php
```diff
@@ -3998,3 +3998,16 @@
 		updateToDbVersion('201805290');
 	}
 }
+
+if (isDatabaseVersion('201805290')) {
+
+	showUpdateStep("Adding leaccount field to panel customers");
+	Database::query("ALTER TABLE `froxlor`.`panel_customers` ADD COLUMN `leaccount` varchar(255) default '' AFTER `leregistered`;");
```
d00p: do not used fixed database name please, just leave the `froxlor`.

d00p:
Could you please explain briefly what the changes - escpecially the leaccount field - are doing and used for?

markuspetermann:
The leaccount field stores the ACME account URL that is received once the account is registered. The ACME specification (https://tools.ietf.org/html/draft-ietf-acme-acme-14) states (p. 12):

"For newAccount requests, [...]. For all other requests, the request is signed using an existing account and there MUST be a "kid" field.  This field MUST contain the account URL received by POSTing to the newAccount resource."

Until now, the the account URL was only available when the account was registered, therefore all subsequent requests failed.

--

The second change fixes a bug when requesting a certificate for more than one subjectAltName (SAN), e.g. example.com / www.example.com. Until now a seperate order was issued for each SAN, which then failed at the end when requesting a certificate for all demains, while the order only contained the last domain.

The correct behaviour would be, to place one order per certificate and process authorization for all included SANs, as described in section 7.4 of the specification.

d00p:
I assume this has been testen thoroughly? ;)

markuspetermann:
Yes, this has been tested, no further issues so far.

PR #576, opened 2018-09-30
Make writing AccessLog and ErrorLog optional

ServiusHack:
In GDPR's spirit of data minimisation it should be possible to not write any log files. With this change each domain gets the option to disable the access and error log of Apache.


d00p:
Nice PR, what about nginx and lighttpd crons?

ServiusHack:
Good question. Didn't think of the others as I'm only using Apache.

Since there we no comments on the code I took the liberty to amend my previous commit. Hope that's ok. Lighttpd and Nginx now also consider the settings. I tested it a bit but not each combination of the settings.

d00p:
not compatible with current master-branch, do you want to adjust the changes?

ServiusHack:
I'll look into it this weekend. Hopefully the pull request won't stay open for another two months then, until there are conflicts again.

d00p:
it stayed open so long because I was about finished with 0.9.x and already putting work in 0.10.0 which is now master branch after the release of 0.9.40

PR #577, opened 2018-10-07
NGINX Reverse Proxy

wohlgemu:
this branch enables froxlor to use NGINX as Reverse Proxy.
The changes are motivated by the external tool published on GitHub repo Ithariel/froxlor_nrp.

It enables the admin to add IP and Ports, which are handled by NGINX.
The Apache Cron-Task automatically creates VHosts having a Proxy-IP using the target IP.

NGINX Vhosts are created for Proxy-IP domains only.

For example creating a SSL IP/Port proxy with IP/Port 1.2.3.4:443 pointing to 1.2.3.4:8443, the following NGINX vHost is created 

```
server {
        listen    1.2.3.4:443 default_server ssl http2;
        # Froxlor default vhost
        server_name    myserver.com;
        access_log      /var/log/nginx/access.log;
        location /.well-known/acme-challenge {
                alias /srv/www/froxlor/.well-known/acme-challenge;
                location ~ /.well-known/acme-challenge/(.*) {
                        default_type text/plain;
                }
        }

        root     /var/www/froxlor/;
        index    index.php index.html index.htm;


        location ~ /\.ht {
                deny all;
        }

        location / {
                try_files $uri $uri/ @proxy;
        }

        location @proxy {
                proxy_pass https://$server_addr:8443$request_uri;
                proxy_set_header Host $host;
                proxy_set_header X-Real-IP $remote_addr;
                proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
                proxy_set_header X-Forwarded-Proto $scheme;
                include /etc/nginx/proxy[.]conf;
        }

        ## This is (for this example) defined as system.default_proxyconf
        location ~ \.(php|cgi|pl|htm?l)$ {
                proxy_pass https://$server_addr:8443$request_uri;
                 include /etc/nginx/proxy[.]conf;
        }
        ssl_protocols TLSv1 TLSv1.2;
        ssl_ciphers ECDH+AESGCM:ECDH+AES256:!aNULL:!MD5:!DSS:!DH:!AES128;
        ssl_ecdh_curve secp384r1;
        ssl_prefer_server_ciphers on;
        ssl_certificate /etc/ssl/froxlor/mydomain.com_bundle.pem;
        ssl_certificate_key /etc/ssl/froxlor/mydomain.com_bundle.key;
}
```

On the the other hand an Apache vHost is created for 1.2.3.4:8443 (as usual), but not for 1.2.3.4:443.


admin_ipsandports.php
```diff
@@ -137,7 +140,10 @@
 					standard_error('cantdeletedefaultip');
 				}
 			} else {
-				standard_error('ipstillhasdomains');
+				if($result_checkdomain['id'] != '')
```
d00p: please use curly braces to enclose the if-statements (like everywhere in the code)

install/froxlor.sql
```diff
@@ -684,8 +685,12 @@ opcache.interned_strings_buffer'),
 	('panel', 'password_special_char_required', '0'),
 	('panel', 'password_special_char', '!?<>§$%+#=@'),
 	('panel', 'customer_hide_options', ''),
-	('panel', 'version', '0.9.39.5'),
-	('panel', 'db_version', '201809280');
+	('system', 'apache_use_nrp', '0'),
+	('system', 'proxyconf_vhost', '/etc/nginx/sites-enabled/')
+        ('system', 'default_proxyconf', '')
+	('system', 'proxyreload_command', 'service nginx reload')
+	('panel', 'version', '0.9.39.6-dev1'),
```
d00p: please do not increase or change version field...only increase db_version (date + n) - also, it is not necessary to append -devX anywhere

install/updates/froxlor/0.9/update_0.9.inc.php
```diff
@@ -3893,6 +3893,12 @@
 	updateToVersion('0.9.39.2');
 }
 
+if (isFroxlorVersion('0.9.39.2-dev1')) {
```
d00p: wrong position, always add changes to the END of the update file. Also, as said earlier, just adjust the db_version and use isDatabaseVersion()

install/updates/froxlor/0.9/update_0.9.inc.php
```diff
@@ -3893,6 +3893,12 @@
...
+
+        showUpdateStep("Updating from 0.9.39.2 to 0.9.39.1-dev1", false);
+        updateToVersion('0.9.39.');
```
d00p: this would be fun for future updates....you are missing "1-dev1" here. But that's wrong anyway as stated already

install/updates/froxlor/0.9/update_0.9.inc.php
```diff
@@ -4060,3 +4065,21 @@
 	
 	updateToDbVersion('201809280');
 }
+
+
+if (isFroxlorVersion('0.9.39.5') || isDatabaseVersion('201809280')) {
```
d00p: only use isDatabaseVersion here

install/updates/froxlor/0.9/update_0.9.inc.php
```diff
@@ -4060,3 +4065,21 @@
...
+
+        showUpdateStep("Adding new fields for NGINX-Reverse-Proxy");
+        Database::query("ALTER TABLE `" . TABLE_PANEL_IPSANDPORTS . "` ADD `proxyto` int(11) NOT NULL DEFAULT '0';");
+        lastStepStatus(0);
+
+        showUpdateStep("Updating settings table");
+        Database::query("INSERT INTO `" . TABLE_PANEL_SETTINGS . "` (`settinggroup`, `varname`, `value`) VALUES ('system', 'apache_use_nrp', '0');");
```
d00p: When adding new Settings, please use Settings::addNew (as used many times in this file)

install/updates/froxlor/0.9/update_0.9.inc.php
```diff
@@ -4060,3 +4065,21 @@
...
+        Database::query("INSERT INTO `" . TABLE_PANEL_SETTINGS . "` (`settinggroup`, `varname`, `value`) VALUES ('system', 'proxyconf_vhost', '');");
+        Database::query("INSERT INTO `" . TABLE_PANEL_SETTINGS . "` (`settinggroup`, `varname`, `value`) VALUES ('system', 'default_proxyconf', '');");
+        Database::query("INSERT INTO `" . TABLE_PANEL_SETTINGS . "` (`settinggroup`, `varname`, `value`) VALUES ('system', 'proxyreload_command', 'service nginx reload');");
+        lastStepStatus(0);
+
+        updateToVersion('0.9.39.6-dev1');
+        updateToDbVersion('201809280-dev1');
```
d00p: db_version is purely numerical...also wrong value

lib/formfields/admin/ipsandports/formfield.ipsandports_add.php
```diff
@@ -16,28 +16,30 @@
  */
 
 return array(
-	'ipsandports_add' => array(
-		'title' => $lng['admin']['ipsandports']['add'],
-		'image' => 'icons/ipsports_add.png',
+	'ipsandports_edit' => array(
```
d00p: did you mix up ipsandports_add and ipsandports_edit??? There is no `$result` when adding new entities

lib/version.inc.php
```diff
@@ -16,10 +16,10 @@
  */
 
 // Main version variable
-$version = '0.9.39.5';
+$version = '0.9.39.6-dev1';
```
d00p: as said, leave the version as it is

lib/version.inc.php
```diff
@@ -16,10 +16,10 @@
...
 
 // Database version (YYYYMMDDC where C is a daily counter)
-$dbversion = '201809280';
+$dbversion = '201809280-dev1';
```
d00p: this should be "201810070" as of today

lng/english.lng.php
```diff
@@ -299,7 +299,6 @@
 $lng['admin']['deactivated'] = 'Deactivated';
 $lng['admin']['deactivated_user'] = 'Deactivate user';
 $lng['admin']['sendpassword'] = 'Send password';
-$lng['admin']['ownvhostsettings'] = 'Own vHost-settings';
```
d00p: Why did you remove that? it's still in use

lng/english.lng.php
```diff
@@ -2127,3 +2126,25 @@
 $lng['admin']['notryfiles']['description'] = 'Say yes here if you want to specify a custom try_files directive in specialsettings (needed for some wordpress plugins for example).';
 $lng['serversettings']['phpfpm_settings']['override_fpmconfig'] = 'Override FPM-daemon settings (pm, max_children, etc.)';
 $lng['serversettings']['phpfpm_settings']['override_fpmconfig_addinfo'] = '<br /><span class="red">Only used if "Override FPM-daemon settings" is set to "Yes"</span>';
+
+
+// FORK
```
d00p: fork? not if it's part of the froxlor-repo - just follow the naming in the file...or better, just append to the end (no unecessary empty lines please)

scripts/jobs/cron_tasks.inc.http.10.apache.php
```diff
@@ -838,12 +843,18 @@ protected function getVhostContent($domain, $ssl_vhost = false)
 		$query = "SELECT * FROM `" . TABLE_PANEL_IPSANDPORTS . "` `i`, `" . TABLE_DOMAINTOIP . "` `dip`
 			WHERE dip.id_domain = :domainid AND i.id = dip.id_ipandports ";
 		
+		// Ignore IPs which are proxied by NRP
```
d00p: This comment would be better placed at line 170 and the following commented code removed (as it is exactly what is used above)

scripts/jobs/cron_tasks.inc.http.10.apache.php
```diff
@@ -882,108 +903,105 @@ protected function getVhostContent($domain, $ssl_vhost = false)
 			}
 			$ipportlist .= $ipport;
 		}
+
```
d00p: what are the massive changes below? They do not seem to have anything to do with IP/Proxy stuff...irritating

templates/Sparkle/admin/ipsandports/ipsandports.tpl
```diff
@@ -32,7 +32,9 @@ $header
 							<th>Specialsettings</th>
 							<if $is_apache><th>ServerName</th></if>
 							<th>SSL</th>
+							<th>NRP</th>
```
d00p: NRP is not really meaningful. Why not just "Reverse proxy"?

scripts/jobs/cron_tasks.inc.http.10.apache.php
wohlgemu: I had to refactor the code a bit because the creating of the vHost code was mixing with setting of properties. I had the issue, that there were vHost container created with an empty $ipportlist, which is now captured later.

lng/english.lng.php
wohlgemu: Was deleted accident

admin_ipsandports.php
wohlgemu: fixed

install/froxlor.sql
wohlgemu: I was not sure how to handle the versioning correctly for branched versions. I hope I fixed it accordingly.

install/updates/froxlor/0.9/update_0.9.inc.php
wohlgemu: removed

d00p:
Hm, what about people using only one IP and add their IP with 443 as proxy and select only 2 of 4 domains to use it, wouldn' the cronjob create 2 entries in nginx listening on ip:443 and 2 entries in apache listening on ip:443? Which then would not work, either one would not start because the it cannot bind to the address/port

d00p:
> Than the cron would create a NGINX vhost domain1.tld for IP:443, a Apache vhost domain1.tld listening > on IP:8443 and a Apache vhost listening on IP:8443 for domain2.tld.

Did you forget that a nginx vhost on IP:443 for domain**2**.tld is also generated because this way, there would be no "normal" vhost for IP:443 for domain2.tld ...only 8443

> With this structure one would not need to change the apache cron, because all ip:port entries are created as usual and only the NGINX vHosts ((proxy_ip|ip):proxy_port) are created, where proxy_port is != '', forwarding to the apache ip:port.

Sounds way easier that way

wohlgemu:
> Sounds way easier that way
I will implement it in this way.

However, I think I have found another problem in generating the vHosts (independent of the reverse proxy)...

If one defines two ip/ports, lets say ip1:port1 and ip2:port2, and assign both IPs to one domain, than the content of the default vHost container of both IPs is included in the vHost, because there is only one vhost having both IPs in the <VirtualHost> created. However, there should be two vHosts: one listening to ip1:port1 and including the default vHost code ip1:port1, and a second vHost listening on ip2:port2 including its default vHost code. In the current implementation the two configurations may conflict. If I am not wrong, I can open an issue.

d00p:
That's a whole other "problem" and has nothing to do with reverse proxy. thing is,this is very specific, not many admins will have to different ip/ports with conflichting custom-vhost content that is being added to the customers-domain-vhost, and out of them there are some that explicitly want it that way.

scripts/jobs/cron_tasks.php
```diff
@@ -84,6 +92,16 @@
 			$webserver->writeConfigs();
 			$webserver->createOwnVhostStarter();
 			$webserver->reload();
+
+			if(1 && $run_nrp) {
```
d00p: unecessary condition if (1) is always true...

```suggestion
			if($run_nrp) {
```

scripts/jobs/cron_tasks.inc.http.40.nginx_proxy.php
```diff
@@ -0,0 +1,1166 @@
+<?php
```
d00p: There is lots of nginx config stuff done that is also done in the nginx-cron...maybe you can override just a view things instead of duplicating? Like we do with apache and apache_fcgid (class apache_fcgid extends apache). Could be possible

install/updates/froxlor/0.9/update_0.9.inc.php
```diff
@@ -4060,3 +4060,33 @@
...
+	Settings::addNew('system.apache_use_nrp', '0');
+        Settings::addNew('system.proxyconf_vhost', '');
+        Settings::addNew('system.default_proxyconf', '');
+        Settings::addNew('system.proxyreload_command', 'service nginx reload');
+        lastStepStatus(0);
+
+        updateToDbVersion('201810070');
+}
+
+if (isDatabaseVersion('201810070')) {
```
d00p: why split this in two database-versions? Unecessary

d00p:
No more work on this happened for a year - close

PR #579, opened 2018-10-20
Add TLSv1.3

quthla:
Confirmed working with Apache2


d00p:
thanks for the changes, but be careful, the current stable apache version does NOT support TLSv1.3 yet and the websites will be inaccessable if acctivated with apache prior 2.4.36

quthla:
That's true. I use 2.4.37 and it's working fine there so I thought I'd PR this.

PR #582, opened 2018-10-24
Support dynamic DNS in Froxlor

ServiusHack:
Customers can update the IP addresses of their domains via a simple HTTP
request.  This must be enabled for the domain before it is available. IPv4 and
IPv6 is supported. Multiple domains in one call are supported.

The `updateip.php` script uses HTTP Basic Authentication. With mod_php and
fcgid no configuration is required but for php-fpm the option '-pass-header
Authorization / CGIPassAuth On' must be enabled for the PHP configuration of
Froxlor's vhost.


ServiusHack:
This has been working for me for several years now. When I first presented it in the forum [1] there wasn't much enthusiasm but I've grown tired of manually patching Froxlor. Hopefully you consider this worth being merged. 

The IPv6 part is not much tested, I'm only using IPv4.

The HTTP authentication is also not essential, most clients support passing credentials as parameters as well. Still, I think it's good to have it.

If you're ok with having this functionality in Froxlor I'm more than happy to make any necessa
ry modifications to get it merged.

[1] https://forum.froxlor.org/index.php?/topic/12790-dynamic-dns-in-froxlor

d00p:
hmmm...basically a good thing yes - i'd rather implement that in the 0.10.0 branch using API so you can simply update DNS records using the API - and I think it's superfluous to mark domains for dynamic-dns and additionally if ipv4 or ipv6.
The customers has domains, he can update all of them if he wants to via API easily - only an updateip-script/wrapper could be provided for easier processing and usage of dyndns protocol.

What do you think?

ServiusHack:
I've only recently heard of the API for 0.10.0. Can you point me to some overall architecture, documentation or other information on how the API will look like?

If 0.10.0 will be the next release I'll happily move this feature to the 0.10.0 branch. If you plan to release another 0.9.x I'd prefer if this could be merged into master.

As you wrote, a wrapper for the dyndns protocol will probably be necessary.

Being able to mark domains for dynamic-dns should be kept, in my opinion. Similar to the limit of subdomains or IMAP this gives the option to sell it as an additional feature to customers. It should also be separate from the full-fledged DNS editor which is more complicated and much more feature-rich. Or am I misunderstanding what you consider to be superfluous?

d00p:
You might be right, for admins the number that increases for used is the number the admin assigns to customers. So what for the customer is assigned, for the admin is used (he gave the resource from his contingent)

updateip.php
```diff
@@ -0,0 +1,197 @@
...
+		header('HTTP/1.1 400 Bad Request');
+		header('Content-Type: text/plain');
+		echo 'Invalid IPv4.';
+		exit;
+	}
+}
+if (isset($_REQUEST['ipv6'])) {
+	$ipv6 = $_REQUEST['ipv6'];
+	if (!filter_var($ipv6, FILTER_VALIDATE_IP, FILTER_FLAG_IPV6)) {
+		header('HTTP/1.1 400 Bad Request');
```
d00p: you're repeating the output-headers very often...why not create a function e.g. like this:
```php
function response($text = null, $code = "400", $codedesc = 'Bad Request') {
   header('HTTP/1.1 '.$code.' '.$codedesc);
   header('Content-Type: text/plain');
   echo $text;
   exit;
}
```

updateip.php
```diff
@@ -0,0 +1,197 @@
...
+
+	$ip = $_SERVER['REMOTE_ADDR'];
+	if (filter_var($ip, FILTER_VALIDATE_IP, FILTER_FLAG_IPV4))
+		$ipv4 = $ip;
+	else if (filter_var($ip, FILTER_VALIDATE_IP, FILTER_FLAG_IPV6))
+		$ipv6 = $ip;
+	else {
+		header('HTTP/1.1 500 Internal Server Error');
+		header('Content-Type: text/plain');
+		echo 'Unable to detect IP. It must be provided.';
```
d00p: why is that an INTERNAL server error? It's a client error for not providing the IP....

lib/Froxlor/Dns/Dns.php
```diff
@@ -217,7 +229,7 @@ public static function createDomainZone($domain_id, $froxlorhostname = false, $i
 										// use the first NS entry as primary ns
 										$primary_ns = $nameserver;
 									}
-									$zonerecords[] = new DnsEntry($record, 'NS', $nameserver);
+									$zonerecords[] = new DnsEntry($record[0], 'NS', $nameserver);
```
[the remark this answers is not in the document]
d00p: Just got it, from `addRequiredEntry()` - I don't really like that....

lib/Froxlor/Api/Commands/Domains.php
```diff
@@ -208,6 +214,11 @@ public function add()
 				$registration_date = $this->getParam('registration_date', true, '');
 				$termination_date = $this->getParam('termination_date', true, '');
 				$caneditdomain = $this->getBoolParam('caneditdomain', true, 0);
+				$isdynamicdomain = $this->getBoolParam('isdynamicdomain', true, 0);
+				$dynamicipv4 = $this->getParam('dynamicipv4', true, '');
+				$dynamicipv4 = ($dynamicipv4 == '') ? null : \Froxlor\Validate\Validate::validate_ip2($dynamicipv4, false, 'invalidip', true, true);
```
d00p: please validate after parameter-listing (original line 233, comment _// validating_))

updateip.php
discordier: Yep, should definitely be `400 Bad request`

d00p: hm, relateable....agreed

ServiusHack:
As discussed in IRC this feature now uses the existing capabilities of Froxlor to change DNS records. This means DNS editor must be enabled. Otherwise the following error message will show:

```
Uncaught Exception: DNS server not enabled on this system in /var/www/froxlor/lib/Froxlor/Api/Commands/DomainZones.php:50
```

There is also no dynamic domains resource anymore. Either a customer can update any of his subdomains (enabled per customer) or he can update a second level domain (enabled per domain).

@d00p please have a look if the change is ok. We can also discuss in IRC again if you want. I'll rebase and fix the merge conflict when you are ok with the proposed changeset.

lib/Froxlor/Cron/Http/Apache.php
```diff
@@ -241,6 +241,7 @@ public function createIpPort()
 						$this->virtualhosts_data[$vhosts_filename] .= '  FcgidIdleTimeout ' . Settings::Get('system.mod_fcgid_idle_timeout') . "\n";
 						if ((int) Settings::Get('system.mod_fcgid_wrapper') == 0) {
 							$this->virtualhosts_data[$vhosts_filename] .= '  SuexecUserGroup "' . Settings::Get('system.mod_fcgid_httpuser') . '" "' . Settings::Get('system.mod_fcgid_httpgroup') . '"' . "\n";
+							$this->virtualhosts_data[$vhosts_filename] .= '  SetEnvIf Authorization .+ HTTP_AUTHORIZATION=$0' . "\n";
```
d00p: is there an equivilant (needed) for Nginx and Lighttpd?

lib/Froxlor/Dns/Dns.php
```diff
@@ -72,6 +72,7 @@ public static function createDomainZone($domain_id, $froxlorhostname = false, $i
 
 		self::addRequiredEntry('@', 'A', $required_entries);
 		self::addRequiredEntry('@', 'AAAA', $required_entries);
+
```
d00p: unnecessary change :)

updateip.php
```diff
@@ -0,0 +1,221 @@
+<?php
```
d00p: Hmm, why so complicated using username/password, re-creating the whole login-process (you dont check for locked users btw).
Just make this script "external" and POST to /api.php using apikeys and call API externally like in the examples here: https://github.com/Froxlor/Froxlor/blob/master/doc/example/list_functions.php

lib/Froxlor/Cron/Http/Apache.php
discordier: AFAIR only apache strips the `Authorization` header when it contains an unknown scheme, so nginx and lighttpd should be unaffected.

ServiusHack: Good question, I've checked all the possible configurations to see if HTTP authentication works. As long as FCGID and PHP-FPM are not used for the Froxlor virtual host everything is fine. If they are used then:

* Apache 2.4 with fcgi needs the suggested change
* Apache 2.4 with PHP-FPM only works if the PHP configuration for the Froxlor VHost has this setting enabled:
> Add "-pass-header Authorization" / "CGIPassAuth On" to vhosts:
* nginx with PHP-FPM works because nginx seems to always pass the HTTP_AUTHORIZATION header to the fastcgi script.
* Nginx with a separate PHP backend works for the same reason, it forwards the header.
* LigHTTPd with neither FCGID nor PHP-FPM gives a configuration that can't do anything with PHP by my understanding. I've ignored this case.
* LigHTTPd with FCGID doesn't support FCGID for the Froxlor vHost according to the UI. I've ignored this case as well.
* LigHTTPd with PHP-FPM also seems to always pass the HTTP_AUTHORIZATION header to the fastcgi script.

Should I somehow make sure that "-pass-header Authorization" / "CGIPassAuth On" is always enabled for the PHP configuration of Froxlors virtual host?

updateip.php
ServiusHack: The script can't be external because it performs authorization. It checks if a customer is allowed to update his subdomains. As discussed in IRC the dynamic DNS feature is not a wrapper of the API but a separate feature. It should be available without allowing the customer to change any DNS entry he likes (like MX, NS, etc.).

Should I refactor the user authentication code to be able to reuse it? Or should I refactor the API authentication code to be able to use it in the script? Or should we talk in IRC?

d00p: Let's better talk on IRC...this might take a while ;)

lib/Froxlor/Cron/Http/Apache.php
discordier: IMO apache 2.4+php-fpm also works without `Add "-pass-header Authorization"` and `CGIPassAuth On`.
I simply have a rewrite rule for this that I stole from Symfony, so far this worked well for any apache setup:
```
    # Sets the HTTP_AUTHORIZATION header removed by Apache
    RewriteCond %{HTTP:Authorization} .
    RewriteRule ^ - [E=HTTP_AUTHORIZATION:%{HTTP:Authorization}]
```
I do not know if this is a proper substitution for your `SetEnvIf` in all cases but I guess so.

d00p:
No more response after discussing on IRC

Issue #584, opened 2018-11-13
PHP 7.2 deprecation warning


Xavjer:
# Summary

After upgrading my server to PHP 7.2, this popped up:

PHP warning/error
#8192 assert(): Calling assert() with a string argument is deprecated
 
/lib/functions/filedir/function.makeCorrectDir.php:32
 
PHP warning/error
#2 Cannot modify header information - headers already sent by (output started at /var/www/froxlor/lib/functions/froxlor/function.phpErrHandler.php:37)
 
/lib/functions/output/function.redirectTo.php:76

Correct as of >>> http://php.net/manual/en/function.assert.php#refsect1-function.assert-parameters

# System information
Debian Jessie
Serversoftware: | Apache/2.4.10 (Debian)
PHP-Version: | 7.2.11-4+0~20181106031710.10+jessie~1.gbp789850
MySQL server version: | 5.5.60-0+deb8u1
Webserver interface: | CGI-FCGI
Memory usage: | MemTotal:        2058316 kB 
MemFree:          197712 kB 
MemAvailable:     887592 kB

Froxlor Version: 0.9.39.5-1+jessie1 (DB: 201802130)

# Steps to reproduce

1. Log in as Admin
2. Go to Resources > Domains
3. Edit a Domain
4. Save => Blank screen with the warnings

# Expected behavior

1. Go back to main page

# Actual behavior

1. Blank screen, even tough it looks as if everything is saved. Need to relogin afterwards

I also got this assert() deprecated error in other places, it is therefore required to search trough all files to find the usages of it.
(For example if logged in as user and go to Domains > Settings, then there are a lot of these messages but they are formatted nicely and the GUI is still there
'''
PHP warning/error
#8192 assert(): Calling assert() with a string argument is deprecated
/lib/functions/filedir/function.makeCorrectDir.php:32
'''

Maybe the error is in other places as well, but I did not find any yet. Also the Cron Job is still running under PHP5.

As written, it looks as if everything is saved therefore no big deal for me at the moment

# Log files/log entries
syslog:
-

d00p:
Please test with latest version and/or git masterAm 13.11.2018 14:51 schrieb Xavjer <notifications@github.com>:Summary
After upgrading my server to PHP 7.2, this popped up:
PHP warning/error
#8192 assert(): Calling assert() with a string argument is deprecated
 
/lib/functions/filedir/function.makeCorrectDir.php:32
 
PHP warning/error
[...]

PR #585, opened 2018-11-13
#584 - PHP deprecated warning

Xavjer:


lib/functions/filedir/function.makeCorrectDir.php
```diff
@@ -27,9 +27,9 @@
 function makeCorrectDir($dir) {
 
 	if (version_compare("5.4.6", PHP_VERSION, ">")) {
-		assert('is_string($dir) && strlen($dir) > 0 /* $dir does not look like an actual folder name */');
```
d00p: the comment was part of the assertion on purpose as there was no description parameter until 5.4.8 and so the comment would be output as well

Xavjer: > the comment was part of the assertion on purpose as there was no description parameter until 5.4.8 and so the comment would be output as well

Ah, that explains why this was in ' all the time. This should therefore be fine now. I tested the else clause on my server and, at least for 7.2 it is now fine. Did not find any errors after that.

d00p: So, if you could change that back for <=5.4.6 I'd gladly merge

d00p:
Argh....next time, please only against MASTER branch...not 0.9.x

Xavjer:
That's why I made the change in 0.9.x ':)
I was unsure if the 0.9.x will be used in the future and therefore wanted to evade that this happens again ;)

d00p:
it will only be used for critical security bugfix releases in the 0.9 series....but we want to focus on the api based 0.10.0 now

PR #594, opened 2018-11-25
Change formfield id for new admins

Broatcast:
little Bug, requested "new_loginname" instead of "loginname"


d00p:
uh yeah, changed that to be the same as in Customers.add() - thx :)

PR #596, opened 2018-11-25
Adds first composer.json file

foliengriller:
If somebody publishs it on packagist, it will allow users to install froxlor via composer. 
I just filled it with the things I considered correct... but maybe I'm wrong...


d00p:
required extensions: 

- session
- ctype
- simplexml
- xml
- filter
- posix
- mbstring
- curl
- json

optional extensions:

- bcmath
- zip

PR #602, opened 2018-11-28
Adds composer.lock and updates .gitignore due to /vendor

foliengriller:


composer.lock
```diff
@@ -9,32 +9,32 @@
...
-                "php": "^7.1"
+                "php": ">=5.3,<8.0-DEV"
             },
             "require-dev": {
                 "athletic/athletic": "~0.1.8",
                 "ext-pdo": "*",
                 "ext-phar": "*",
-                "phpunit/phpunit": "^6.2.3",
-                "squizlabs/php_codesniffer": "^3.0.2"
+                "phpunit/phpunit": "~4.0",
```
d00p: tests require phpunit 6

d00p:
Don't know why this should be needed....composer.json and vendor/* are not used in production...only for testing

composer.lock
foliengriller: This is not for production. You set up a CI-Pipeline. That requires some stuff from composer. Right? 
If you do not commit the composer.lock, you can not ensure, that it works like expected. See here: https://getcomposer.org/doc/01-basic-usage.md#commit-your-composer-lock-file-to-version-control 

You worked around this problem with very specific version requirements in composer.json. This is not how composer is supposed to work. 
The exact version etc. of all needed packages is written in composer.lock. This file is automatically created by composer and you do not need to fullfill the dev-requirements of all the packages it contains... It only changes if you update a package or change something in the composer.json.

d00p: I know what a composer.lock file is thank you. Also I don't require these exact version but I wanted to be sure that the tests pass in travis like they do in Jenkins. Still don't see the need to have the composer.lock in the repo nor the exclusion of vendor/ as it is never part of froxlor only for testing

d00p: > The travis task does a composer install, right?

yes, because the tools use in the ant-build.xml are not pre-installed

> So due to the missing of a composer.lock it will fallback to your composer .json. Thas why you added there a lot of very specific versions (so composer is quite useless):

No, this is not why I've added specific versions there, I already wrote why the versions are so specific. And no, composer is not useless, because it installes the tools required for the tests on the travis-test-server.

> But this will NOT ensure, that allways exactly the same packages are used, what is importend for the test-environment

The only REALLY important thing is phpunit being version 6 (which then requires php-7) and nothing else...all the other tools are just added because the ant-build-script includes all these from the jenkins-CI - and there they are all installed (that's why there never was a composer.json before).

d00p: froxlor does not support adding stuff via composer...we have our own autoloader etc. that would mean to completely restructure the classes e.g. - it's not that I don't know how composer works and that it is a great tool but we do not use it - only for our unit-tests.

To answer your questions: IF everything was set up to use external libraries via vendor/ directory, THEN if course a composer.lock makes sense

foliengriller: Ok, so it is just to early...

d00p: Definetly...when there's time for a new frontend I will for sure use external classes via composer...but this is a huge refactor process I currently cannot handle

PR #609, opened 2018-12-04
Improves annotations in /lib/classes/api/

foliengriller:


lib/classes/api/class.FroxlorRPC.php
```diff
@@ -79,10 +79,10 @@ private static function validateAuth($key, $secret)
 	/**
 	 * validates the given command
 	 *
-	 * @param array $body
+	 * @param $request
```
d00p: variable type missing

lib/classes/api/abstract.ApiCommand.php
```diff
@@ -285,8 +287,7 @@ protected function mailer()
 	/**
 	 * call an api-command internally
 	 *
-	 * @param string $module
-	 * @param string $function
+	 * @param null $command
```
d00p: $command is a "string", not "null"

lib/classes/api/abstract.ApiParameter.php
```diff
@@ -145,6 +146,9 @@ protected function getParamList()
 	 * @param int $level
 	 *        	depth of backtrace, default 2
 	 *        	
+	 * @param int $max_level
+	 * @param null $trace
```
d00p: $trace is array|null

PR #610, opened 2018-12-04
Corrects annotations in class.Admins.php

foliengriller:


lib/classes/api/commands/class.Admins.php
```diff
@@ -23,7 +23,7 @@ class Admins extends ApiCommand implements ResourceEntity
 	 *
 	 * @access admin
 	 * @throws Exception
-	 * @return array count|list
+	 * @return string
```
d00p: basically true, a json_encoded array is returned (every api call does) but the extra info should be kept because the phpdoc is parsed for the api-documentation at https://api.froxlor.org/doc/0.10.0/).
So maybe we can agree on

`@return string json-encoded array count|list`?

lib/classes/api/commands/class.Admins.php
```diff
@@ -57,7 +57,7 @@ public function listing()
 	 *        	
 	 * @access admin
 	 * @throws Exception
-	 * @return array
+	 * @return string
```
d00p: see listing()

lib/classes/api/commands/class.Admins.php
```diff
@@ -160,7 +160,7 @@ public function get()
...
```
d00p: see listing() but not with count|list

lib/classes/api/commands/class.Admins.php
```diff
@@ -437,7 +437,7 @@ public function add()
...
```
d00p: see listing() but not with count|list

lib/classes/api/commands/class.Admins.php
```diff
@@ -694,7 +694,7 @@ public function update()
...
```
d00p: see listing() but not with count|list

lib/classes/api/commands/class.Admins.php
```diff
@@ -794,7 +794,7 @@ public function delete()
...
```
d00p: see listing() but not with count|list

d00p:
Also, this would apply to almost all API calls

d00p:
> This are not really parameters of this method. I know, they come via $this->getParam. But it looks horrible to me and is also not that easy to maintain.
> Is it possible to build a object out of the params first and pass this to the method?

Surely possible, but I won't change that now, don't want functions with these amount of parameters, no fun to code. Also, the parameters documentation would stay the same

d00p:
Additional: if you want to discuss this more deeply, best join us on IRC (freenode #froxlor)

lib/classes/api/commands/class.Admins.php
foliengriller: Thats ok for me...

foliengriller:
Did I still miss something?

PR #612, opened 2018-12-12
Checks Mysql version only if no fatal error occured

foliengriller:
Otherwise it is not possible to install current master due to fatal error

` PHP Fatal error:  Uncaught Error: Non-static method PDO::getAttribute() cannot be called statically in /var/www/html/froxlor/install/lib/class.FroxlorInstall.php:534`

### DIFF
--- a/install/lib/class.FroxlorInstall.php
+++ b/install/lib/class.FroxlorInstall.php
@@ -234,14 +234,14 @@ class FroxlorInstall
 				$fatal_fail = true;
 			}
 		}
-		$version_server = $db_root->getAttribute(PDO::ATTR_SERVER_VERSION);
-		$sql_mode = 'NO_ENGINE_SUBSTITUTION';
-		if (version_compare($version_server, '8.0.11', '<')) {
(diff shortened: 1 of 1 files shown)


install/lib/class.FroxlorInstall.php
```diff
@@ -234,14 +234,14 @@ private function _doInstall()
...
 		}
-		$version_server = $db_root->getAttribute(PDO::ATTR_SERVER_VERSION);
-		$sql_mode = 'NO_ENGINE_SUBSTITUTION';
-		if (version_compare($version_server, '8.0.11', '<')) {
-			$sql_mode .= ',NO_AUTO_CREATE_USER';
-		}
-		$db_root->exec('SET sql_mode = "'.$sql_mode.'"');
 		
 		if (! $fatal_fail) {
+			$version_server = $db_root->getAttribute(PDO::ATTR_SERVER_VERSION);
```
d00p: the move is correct yes...settings the mode only makes sense if connection works

PR #613, opened 2018-12-13
Uses more single quotes

foliengriller:
I have tested the install-process once on debian 9 with apache 2.4, without any problems.

### DIFF
--- a/install/lib/class.FroxlorInstall.php
+++ b/install/lib/class.FroxlorInstall.php
@@ -205,7 +205,7 @@ class FroxlorInstall
 	 */
 	private function _doInstall()
 	{
-		$content = "<table class=\"noborder\">";
+		$content = '<table class="noborder">';
 		
 		// check for mysql-root-connection
 		$content .= $this->_status_message('begin', $this->_lng['install']['testing_mysql']);
@@ -244,7 +244,7 @@ class FroxlorInstall
 			$db_root->exec('SET sql_mode = "'.$sql_mode.'"');
 			
 			// ok, if we are here, the database connection is up and running
-			$content .= $this->_status_message('green', "OK");
+			$content .= $this->_status_message('green', 'OK' );
 			// check for existing db and create backup if so
 			$content .= $this->_backupExistingDatabase($db_root);
 			// create unprivileged user and the database itself
@@ -284,7 +284,7 @@ class FroxlorInstall
 			}
 		}
 		
-		$content .= "</table>";
+		$content .= '</table>';
 		
 		// check if we have unrecoverable errors
 		if ($fatal_fail || $another_fail || $this->_abort) {
@@ -316,7 +316,7 @@ class FroxlorInstall
 	 */
 	private function _createUserdataConf()
 	{
-		$content = "";
+		$content = '';
 		
 		$content .= $this->_status_message('begin', $this->_lng['install']['creating_configfile']);
 		$userdata = "<?php\n";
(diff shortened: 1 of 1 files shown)


install/lib/class.FroxlorInstall.php
```diff
@@ -53,23 +53,23 @@ class FroxlorInstall
...
 	 * @var array
 	 */
-	private $_data = null;
+	private $_data;
 
 	/**
 	 * supported languages for install
 	 */
-	private $_languages = array(
+	private $_languages = [
```
d00p: unnecessary and it's not the froxlor way...we use array() everywhere

install/lib/class.FroxlorInstall.php
```diff
@@ -39,7 +39,7 @@ class FroxlorInstall
 	 *
 	 * @var string
 	 */
-	private $_basepath = null;
+	private $_basepath;
```
d00p: why not initialize variables? unnecessary

install/lib/class.FroxlorInstall.php
```diff
@@ -90,8 +90,8 @@ class FroxlorInstall
 	 */
 	public function __construct()
 	{
-		$this->_basepath = dirname(dirname(dirname(__FILE__)));
-		$this->_data = array();
+		$this->_basepath = dirname(dirname( __DIR__ ));
```
d00p: yes, makes sense

install/lib/class.FroxlorInstall.php
```diff
@@ -102,7 +102,7 @@ public function run()
 		// send headers
 		$this->_sendHeaders();
 		// check if we have a valid installation already
-		$this->_checkUserdataFile();
+		$this->_checkUserDataFile();
```
d00p: unnecessary

install/lib/class.FroxlorInstall.php
```diff
@@ -205,15 +205,15 @@ private function _checkPostData()
...
-		$content = "<table class=\"noborder\">";
+		$content = '<table class="noborder">';
 		
 		// check for mysql-root-connection
 		$content .= $this->_status_message('begin', $this->_lng['install']['testing_mysql']);
 
 		$options = array(
 			'PDO::MYSQL_ATTR_INIT_COMMAND' => 'SET names utf8'
 		);
-		$dsn = "mysql:host=" . $this->_data['mysql_host'] . ";";
```
d00p: unnecessary

install/lib/class.FroxlorInstall.php
```diff
@@ -222,9 +222,9 @@ private function _doInstall()
 			try {
 				$db_root = new PDO($dsn, $this->_data['mysql_root_user'], '', $options);
 				// set the given password
-				$passwd_stmt = $db_root->prepare("
+				$passwd_stmt = $db_root->prepare( '
```
d00p: all queries are enclosed in " "  - and not single quotes

d00p:
Okay, so may unnecessary changes; I think it only makes sense where quotes are quoted in a string, nowhere else

d00p:
wanted to add: i do not try to avoid short array syntax, it's there since php-5.4  but if you start that in one file, you have to update all files thats why i said, better not

foliengriller:
Now I just commited the simple cases. This imho leads to more consistency in the code...

d00p:
Consistency in only one file is meaningless

PR #616, opened 2018-12-20
Unifies type casting via intval() to (int) and floatval() to (float)

foliengriller:
Changes are done via PHP-Storm-inspection. I have looked at it and it looks good to me.


d00p:
Hey thanks for all the work. It would be helpful if you could join us on IRC, so we can talk before you do so many changes as there is going to be huge change soon regarding file structure and OOP

d00p:
Too many conflicts for no real enhancement/advantages

PR #631, opened 2019-01-07
Added entry for 'ftp' to the DNS

princeofnaxos:
# Description

An entry for 'ftp' fails in the DNS zone file. 

## Type of change

# How Has This Been Tested?

Ran the master cronjob.

**Test Configuration**:

* Distribution: Debian 9
* Webserver: Apache 2.4
* PHP: PHP-FPM 7.3

# Checklist:


d00p:
don't see why this should be necessary

d00p:
at least make it an option as we do it for mail-entries (setting system.dns_createmailentry)

princeofnaxos:
Because ftp.[domain] doesn't exist, when not a wildcard domain? 
Customers that use FTP are used to that subdomain.

d00p:
Then please make it an option.

discordier:
And please default it to off!

I do not want my zone files cluttered with hosts that are not needed (my FTP is running on one IP whereas the HTTP is running on another) so this would definitely cause confusion for many customers why there is a ftp defined but not reachable.

PR #635, opened 2019-01-13
htpasswd only on configured paths

heavygale:
Only apply htpasswd rules to (sub)domains if the rule's path begins with the domain's document root.


d00p:
We've talked about this yesterday, makes sense :)

PR #657, opened 2019-02-17
Run acme.sh upgrade just once when cron starts

patschi:
Only run acme.sh upgrade command once when cronjob starts, and not everytime when there's a certificate going to be renewed.

### DIFF
--- a/lib/Froxlor/Cron/Http/LetsEncrypt/AcmeSh.php
+++ b/lib/Froxlor/Cron/Http/LetsEncrypt/AcmeSh.php
@@ -32,6 +32,7 @@ class AcmeSh extends \Froxlor\Cron\FroxlorCron
 	public static function run()
 	{
 		self::checkInstall();
+		self::checkUpgrade();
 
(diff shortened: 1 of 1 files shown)


d00p:
ah, didn't know it automatically checks for updates everytime...thanks...if only i hadn't just changed a lot of code xD

d00p:
Hm, seems like acme.sh is downloading nad installing regardless of the version...maybe we should checkUpdate only once a day or something....it even does this when there's nothing to renew or similar, meaning every 5 minutes (default)....

patschi:
What the idea of doing the `checkUpgrade()` just before running it the first renewal command in the cronjob AND when there's at least one certificate to issue or renew by Froxlor? This way we don't need an own cronjob or so for that.

`acme.sh` itself is quite small and the download is quite fast. So we ensure that the client is always up-to-date prior renewing, in case of any fixes or improvements or so.

d00p:
just commited something xD but yes, that idea is better :)

PR #679, opened 2019-05-03
#564 Allow CIDR and Netmask in mysql_host_access 

### LINKED ISSUE #564: Allow Subnet in MySQL-Access-Hosts
# Summary
In some setups it might be nessesery to grant not only some ips access to the database but a whole subnet. To allow this in mysql you net to set the network with the netmask like 100.127.0.0/255.255.0.0 as host. It is possible to configure this manuall over the command line but the froxlor webinterface das not allow this subnet definition in the settings for MySQL-Access-Hosts.

# System information
* Froxlor version: 0.9.39.5
* Web server: apache2
* DNS server: PowerDNS (standalone)
* POP/IMAP server: Dovecot
* SMTP server: postfix
* FTP server: proftpd
* OS/Version: Debian 9

pquerner:
# Description

In #564 its asked to allow CIDR and netmask of mysql host net validation and saving.

Fixes #564

## Type of change

- [x] This change requires a documentation update **(Possibly. I've changed description field and it should be clear from there to admins.) (Only english and german)**

# How Has This Been Tested?

Locally, with cidr and netmask syntax. Values: 172.23.0.5,172.23.0.5,100.127.0.0/255.255.0.0,10.10.10.10/16,1.1.1.1/16,1.1.1.1/4

(which becomes 172.23.0.5,172.23.0.5,100.127.0.0/255.255.0.0,10.10.10.10/255.255.0.0,1.1.1.1/255.255.0.0,1.1.1.1/240.0.0.0)

CIDR is converted to netmask at saving process, since MySQL allows netmask syntax only.

- [x] Added values in backend (admin account)
- [x] Checked user privilages via mysql workbench. The values are positivly validated.
- [x] Checked if subnet validation is allowed by mysql, by trying to login with a test user with a specific subnet via phpmyadmin.

**Test Configuration**:

* Distribution: Docker (ddev)
* Webserver: nginx/1.14.2
* PHP: 7.2.16-1+0~20190307202415.17+stretch~1.gbpa7be82
* Tested against 0.10.0-rc1 (DB: 201904250)

# Checklist:

- [x] I have commented my code, particularly in hard-to-understand areas

I've only touched english and german localization, since I do not speak other languages. Therefore, for completions sake, we should ask native speakers to translate the string in `$lng['serversettings']['mysql_access_host']['description']` .


pquerner:
@d00p  is there anything else I should change or is missing?

d00p:
IPv6?

pquerner:
According to docs netmask notation cannot be used for IPv6 addresses.

https://dev.mysql.com/doc/refman/5.7/en/account-names.html

d00p:
> A host value can be a host name or an IP address (IPv4 or IPv6).

So,  ipv6 addresses per se are allowed, just not the CIDR notation, does you code cover that? A very nice addon would be a phpunit tests that validates your changes if possible (maybe add a `SettingsStoreTest` to  `tests/Froxlor/`)

d00p:
No more response?

pquerner:
Sorry I have a lot on my plate. If this PR is still interesting you can leave it open and I'll take care of it. Otherwise you can close it.

makuser:
While you are on it, maybe you could change the validation, to allow for %, which is a wildcard in this syntax, and that's currently not allowed.

d00p:
> While you are on it, maybe you could change the validation, to allow for %, which is a wildcard in this syntax, and that's currently not allowed.

would that comply with the add-user / grant stuff? never seend `create user 'someuser'@'%' ...`

makuser:
Well that's exactly what I need I for. I currently need to set it manually...
Since I'm running a MySQL cluster that will be accessed via haproxy, servers will be accessed via their FQDN.

d00p:
> Well that's exactly what I need I for. I currently need to set it manually...
> Since I'm running a MySQL cluster that will be accessed via haproxy, servers will be accessed via their FQDN.

I didn't say we couldn't do that, just wanted to know upfront if just allowing `%` will break other things :)

makuser:
Wouldn't suggest it if it break things, right?😂

pquerner:
Yes, I will actually switch to https://github.com/mlocati/ip-lib for validation and converting. Its test proven and actually a nice little library for this usage.

d00p:
Okay, looking forward to it

d00p:
Anything new here?

pquerner:
Not sure anymore whats the TODO here. The solution provided should still stand, however I have not tested it since working on it the day I provided the PR. Can we split the work perhaps and do the library integration in a seperate issue and the % Wildcard too? Would that work?

lib/Froxlor/Settings/Store.php
```diff
@@ -290,6 +290,28 @@ public static function storeSettingIpAddress($fieldname, $fielddata, $newfieldva
...
+            $ips = explode(',', $ips);
+        }
+        if(is_array($ips) && count($ips) > 0) {
+            $newfieldvalue = [];
+            foreach ($ips as $ip) {
+                $org_ip = $ip;
+                $ip_cidr = explode("/", $ip);
+                if (count($ip_cidr) == 2) {
+                    $ip = $ip_cidr[0];
+                    if(in_array((int)strlen((string)$ip_cidr[1]),array(1,2))) {
```
d00p: this looks really ugly... why not just do `if (strlen($ip_cidr[1]) <= 2) {` as it can not be 0 due to the check on line 303 if there are 2 array elements

lib/Froxlor/Settings/Store.php
```diff
@@ -290,6 +290,28 @@ public static function storeSettingIpAddress($fieldname, $fielddata, $newfieldva
...
 	{
+        $ips = $newfieldvalue;
+	    //Convert cidr to netmask for mysql, if needed be
+        if(strpos($ips, ',') !== false) {
+            $ips = explode(',', $ips);
+        }
+        if(is_array($ips) && count($ips) > 0) {
+            $newfieldvalue = [];
+            foreach ($ips as $ip) {
+                $org_ip = $ip;
```
d00p: I would suggest checking for ipv6 address here and not performing the `cidr2NetmaskAddr` validate etc. on it as it's not valid for MySQL to specify ipv6-cidr's

lib/Froxlor/Validate/Validate.php
```diff
@@ -64,6 +64,26 @@ public static function validate($str, $fieldname, $pattern = '', $lng = '', $emp
...
+    public static function cidr2NetmaskAddr ($cidr) {
+
+        $ta = substr ($cidr, strpos ($cidr, '/') + 1) * 1;
+        $netmask = str_split (str_pad (str_pad ('', $ta, '1'), 32, '0'), 8);
+
+        foreach ($netmask as &$element) {
+            $element = bindec ($element);
+        }
+
+        return join ('.', $netmask);
```
d00p: `join` is just an alias for `implode`, please use `implode` because 

- we always use `implode`
- you never know when an alias might get removed from php in later versions

lib/Froxlor/Validate/Validate.php
```diff
@@ -90,6 +110,9 @@ public static function validate_ip2($ip, $return_bool = false, $lng = 'invalidip
 			$ip_cidr = explode("/", $ip);
 			if (count($ip_cidr) == 2) {
 				$ip = $ip_cidr[0];
+				if(in_array((int)strlen((string)$ip_cidr[1]),array(1,2))) {
```
d00p: same as in lib/Froxlor/Settings/Store.php:305 - better use `if (strlen($ip_cidr[1]) <= 2) {` and check for ipv6

d00p:
Also, the unit-tests fail as you've changed the behavior of `Validate::validate_ip2()` - could you please only convert CIDR to netmask when validating mysql-hosts to avoid breakage of other functionality

pquerner:
How would I know from where I got called from ?
Do you have some solution regarding session or state ?

lib/Froxlor/Settings/Store.php
pquerner: True, I can change that.

lib/Froxlor/Validate/Validate.php
pquerner: Changed.

lib/Froxlor/Validate/Validate.php
pquerner: Changed

lib/Froxlor/Settings/Store.php
pquerner: I must disagree. Checks are done before hand (`\Froxlor\Validate\Check::checkMysqlAccessHost`) and if we're not validating here but in the save method, what good does that do? It just shifts the problem into a new class instead.

d00p:
> Is it OK to introduce a new parameter for this ? Do you see a different way ?

I think that would be the best way, right after `$allow_cidr = false`let's add `$cidr_as_netmask = false` (attention when the last parameter `$throw_exception` is set!!! Adjust accordingly if present) and add the corresponding parameter to your check routine in `Validate\Check::checkMysqlAccessHost()`

lib/Froxlor/Settings/Store.php
d00p: Yes, this needs to be done in `Validate\Check::checkMysqlAccessHost()`.

If we'd do that in `Validate::validate_ip2()` it would be misleading that an IPv6 cidr-notation is invalid which is generally not, only in case of MySQL.

pquerner:
Please check it now. The new introduced paramter is only set to true from `\Froxlor\Validate\Check::checkMysqlAccessHost` so far, so I think its ok now the way it works.
I've fixed the unit tests and added new ones for ipv6 addresses.

lib/Froxlor/Validate/Validate.php
```diff
@@ -62,32 +62,77 @@ public static function validate($str, $fieldname, $pattern = '', $lng = '', $emp
...
+
+    /**
+     * Checks if an $address (IP) is IPv6
+     *
+     * @param $address
+     *
+     * @return bool
+     * @thx https://stackoverflow.com/a/13677565/3020926
+     */
+    public static function is_ipv6($address) {
```
d00p: Don't understand why this should be needed, there's `filter_var()` for checking whether an IP address is ipv6 or not

lib/Froxlor/Validate/Validate.php
```diff
@@ -62,32 +62,77 @@ public static function validate($str, $fieldname, $pattern = '', $lng = '', $emp
...
+     * @param bool   $allow_priv
+     *            whether to allow private network addresses
+     * @param bool   $allow_cidr
+     *            whether to allow CIDR values e.g. 10.10.10.10/16
+     * @param bool   $cidr_as_netmask
+     *            whether to format CIDR nodation to netmask notation
+     *
+     * @return string|bool ip address on success, false on failure
+     */
+	public static function validate_ip2($ip, $return_bool = false, $lng = 'invalidip', $allow_localhost = false, $allow_priv = false, $allow_cidr = false, $throw_exception = false, $cidr_as_netmask = false)
```
d00p: please make `$throw_exception` the last parameter as I've mentioned before (it's always the last parameter also on other functions, it's irritating if this is not the last parameter)

tests/Froxlor/ValidateTest.php
```diff
@@ -86,12 +86,29 @@ public function testValidateIpCidr()
 		$this->assertEquals("12.34.56.78/24", $result);
 	}
 
+    public function testValidateIpv6Disallowed()
+    {
+        $this->expectException("Exception");
+        $this->expectExceptionCode(400);
+        $result = Validate::validate_ip2("2620:0:2d0:200::7/32", false, 'invalidip', false, false, true, true);
```
d00p: you don't need `$result =` if you expect an exception

tests/Froxlor/ValidateTest.php
```diff
@@ -86,12 +86,29 @@ public function testValidateIpCidr()
 		$this->assertEquals("12.34.56.78/24", $result);
 	}
 
+    public function testValidateIpv6Disallowed()
+    {
+        $this->expectException("Exception");
```
d00p: Please expect a certain Message, only expecting `Exception` is very vague

lib/Froxlor/Validate/Validate.php
```diff
@@ -62,32 +62,77 @@ public static function validate($str, $fieldname, $pattern = '', $lng = '', $emp
...
 	{
 		$cidr = "";
 		if ($allow_cidr) {
 			$org_ip = $ip;
 			$ip_cidr = explode("/", $ip);
-			if (count($ip_cidr) == 2) {
+			if (count($ip_cidr) === 2) {
```
d00p: why changing `==` to `===` ....`count` returns `int` and `2` is an integer....a bit unnecessary imho

lib/Froxlor/Validate/Validate.php
```diff
@@ -62,32 +62,77 @@ public static function validate($str, $fieldname, $pattern = '', $lng = '', $emp
...
+			    if (self::is_ipv6($ip)) {
```
d00p: you call somethin `is_ipv6()` and then validate whether it's ipv6 using `filter_var`....seems odd to me

lib/Froxlor/Validate/Validate.php
```diff
@@ -62,32 +62,77 @@ public static function validate($str, $fieldname, $pattern = '', $lng = '', $emp
...
+                    //MySQL does not handle CIDR of IPv6 addresses, return error
+                    if (false === filter_var($ip, FILTER_VALIDATE_IP, FILTER_FLAG_IPV6)) {
```
d00p: please compare function-call to const/check-value, not the other way around, we never do that (as far as I'm aware)

lib/Froxlor/Validate/Validate.php
pquerner: That means every call needs to be changed. Why is this the proper solution here?

tests/Froxlor/ValidateTest.php
pquerner: Will remove it...

lib/Froxlor/Validate/Validate.php
pquerner: So the IDE does not scream anymore at this.

lib/Froxlor/Validate/Validate.php
pquerner: First I check if its any type of v6, then I check (with `ip_cidr`) if it contained `/` which means its a ipv6 cidr notation. Thats a no go here, ipv6 without notation is just fine. I dont see any issue here.

lib/Froxlor/Validate/Validate.php
pquerner: Changed.

lib/Froxlor/Validate/Validate.php
d00p: I thought I already explained why: it's always the last parameter also on other functions, it's irritating if this is not the last parameter

The `$throw_exception` was introduced to give a correct API-response when used within the API and remain behaviour if used in Frontend.

Also, it's only 8 occurences...and I bet not even half of them need to be adjusted - I just like to keep things consistent

lib/Froxlor/Validate/Validate.php
d00p: mine's not :P well, so be it, it's not wrong and can't hurt

tests/Froxlor/ValidateTest.php
pquerner: So .. unlike here?! https://github.com/Froxlor/Froxlor/blob/0afbe3d13b67d439177ae23a4f60028b91cc7687/tests/Froxlor/ValidateTest.php#L72

lib/Froxlor/Validate/Validate.php
d00p: well the "issue" is, why have a function `is_ipv6()` when there's `filter_var($ip, FILTER_VALIDATE_IP, FILTER_FLAG_IPV6)` - hence my question what's up with the `is_ipv6()` function

lib/Froxlor/Validate/Validate.php
pquerner: Instead of thinking around it in another topic regarding this (introducing new parameters often comes with such questions and can easily be dealt with somewhere else) you make everyone adapt to your quirks. Thats why I question everything here. But w/e, I'll change it.

tests/Froxlor/ValidateTest.php
d00p: Hm must be a reason why I did not specify the exceptionmessage (like in any other test...) cant remember. Let's leave it like that :)

lib/Froxlor/Validate/Validate.php
d00p: thank you

lib/Froxlor/Validate/Validate.php
pquerner: Because `filter_var($ip, FILTER_VALIDATE_IP, FILTER_FLAG_IPV4)` reports false, because of the CIDR notation. And if ipv4 is checked with the IPv6 flag, guess what happens.

d00p: Ugh? At that point, there's no need for the CIDR notation for validation, you're exploding `$ip_cidr` and can use `$ip_cidr[0]` or am I missing something?

pquerner: Brainfart by me. Removed the inner If-Statement. My bad.

lib/Froxlor/Validate/Validate.php
pquerner: I'll leave that because there is something called mapped ipv6 address. Like ::ffff:10.20.30.40 which translates to a ipv4 address but in ipv6 "language". If I understand that correctly, tho. That function takes care of it, while `filter_var` apparanetly does not (Did not test!).

d00p: Yeah I've read that too...I would drop that for now because if you support that here you'd have to support this notation everywhere

pquerner: `filter_var($address, FILTER_VALIDATE_IP, FILTER_FLAG_IPV6)` does return false when `$address` is `2620:0:2d0:200::7/32.

lib/Froxlor/Validate/Validate.php
pquerner: Changed it https://github.com/pquerner/Froxlor/commit/9689afc759e1e79a1c10f28afb8b0eab0e6c3a49

lib/Froxlor/Validate/Validate.php
d00p: still don't get the point why you need that - you have the `$ip_cidr` array at the point where you call `self::is_ipv6()` (iirc) which for my understanding should be:
```php
$ip_cidr = [
    0 => 2620:0:2d0:200::7,
    1 => 32
];
```
So, just use `filter_var` on `$ip_cidr[0]`

pquerner: Changed.

lib/Froxlor/Settings/Store.php
pquerner: Added a new parameter for this, so the checks can stay in the validate method.

lib/Froxlor/Validate/Validate.php
```diff
@@ -62,32 +62,78 @@ public static function validate($str, $fieldname, $pattern = '', $lng = '', $emp
...
 		$cidr = "";
 		if ($allow_cidr) {
 			$org_ip = $ip;
 			$ip_cidr = explode("/", $ip);
-			if (count($ip_cidr) == 2) {
+			if (count($ip_cidr) === 2) {
+			    if(strlen($ip_cidr[1]) <= 2 && in_array((int)$ip_cidr[1], array_values(range(1, 32)), TRUE) === false) {
+                    \Froxlor\UI\Response::standard_error($lng, $ip, $throw_exception);
+                }
+                if ($cidr_as_netmask && self::is_ipv6($ip_cidr[0])) {
```
d00p: now there's `is_ipv6()` for only one occurence, might as well do this and remove `is_ipv6()` completely 
```diff
- if ($cidr_as_netmask && self::is_ipv6($ip_cidr[0])) {
+ if ($cidr_as_netmask && filter_var($ip_cidr[0], FILTER_VALIDATE_IP, FILTER_FLAG_IPV6)) {
```

pquerner: Maybe I can add some unit tests for that. Would that help? I think its best if we have this capsuled in case its needed sometime down the road.

d00p: It's used in various places, could be helpful to have that capsuled but then use it everywhere where `filter_var($xyz, FILTER_VALIDATE_IP, FILTER_FLAG_IPV6)` is used (counting 18 occurences using grep)

pquerner: We can change that. But lets tackle this in a new PR.

d00p: agreed

pquerner:
Can you check if this test works?

```
public function testIPv6Checks()
    {
        $result = Validate::is_ipv6('1.1.1.1');
        $this->assertFalse($result);
        $result = Validate::is_ipv6('::ffff:10.20.30.40');
        $this->assertEquals('::ffff:10.20.30.40', $result);
        $result = Validate::is_ipv6('2620:0:2d0:200::7/32');
        $this->assertFalse($result);
        $result = Validate::is_ipv6('2620:0:2d0:200::7');
        $this->assertEquals('2620:0:2d0:200::7', $result);
    }
```

`filter_var` returns the data filtered (ie. string) or false on failure. so I think the code I've used in the if statements are wrong. Can you confirm ?

d00p:
> Can you check if this test works?
> 
> ```
> public function testIPv6Checks()
>     {
>         $result = Validate::is_ipv6('1.1.1.1');
>         $this->assertFalse($result);
>         $result = Validate::is_ipv6('::ffff:10.20.30.40');
>         $this->assertEquals('::ffff:10.20.30.40', $result);
>         $result = Validate::is_ipv6('2620:0:2d0:200::7/32');
>         $this->assertFalse($result);
>         $result = Validate::is_ipv6('2620:0:2d0:200::7');
>         $this->assertEquals('2620:0:2d0:200::7', $result);
>     }
> ```
> 
> `filter_var` returns the data filtered (ie. string) or false on failure. so I think the code I've used in the if statements are wrong. Can you confirm ?

Test-function passes fully

PR #680, opened 2019-05-10
Feature-request #672 - database name prefixes + custom name

pquerner:
Add mysql prefix option to use customername + database description field as final databasename

Fixes #672 

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] This change requires a documentation update

# How Has This Been Tested?

Locally. If the newely option value "DBNAME" is used, new functionality is triggered and executed.

# Checklist:

- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes

I've only added english and german localization, as I do not speak the other languages.

PR #681, opened 2019-05-10
Fix db param null

pquerner:
# Description

Via PR #680 I've seen the method \Froxlor\Database\Manager\DbManagerMySQL::createDatabase. But it allowed "NULL" as parameter value and it will crash the following query (CREATE DATABASE ...). 
Therefore I've added checks to prevent that behaviour and instead throw a exception.

## Type of change

# Checklist:


pquerner:
If it never takes null, why make it optional then by definition?

d00p:
No particular reason ;)

PR #689, opened 2019-05-17
* Make ftp quota related to used traffic instead of used disk space

realbugi:
# Description

Currently Froxlor calculates FTP quota based on disk space usage which is wrong. If we setup traffic (bandwidth) limit lower than disk space in current state then this prevents user from FTP usage. 

Fixes # (issue)

## Type of change

# How Has This Been Tested?

Checking working instance after few days and current system calculation manually.

- [ ] Test A
- [ ] Test B

**Test Configuration**:

* Distribution:
* Webserver:
* PHP:
* etc.etc.:

# Checklist:

- [ ] I have performed a self-review of my own code

### DIFF
--- a/lib/Froxlor/Cron/Traffic/TrafficCron.php
+++ b/lib/Froxlor/Cron/Traffic/TrafficCron.php
@@ -477,7 +477,7 @@ class TrafficCron extends \Froxlor\Cron\FroxlorCron
 			 * Proftpd Quota
 			 */
 			$upd_data = array(
-				'biu' => ($current_diskspace['all'] * 1024),
+				'biu' => ($sum_month_traffic['all'] * 1024),
 				'loginname' => $row['loginname'],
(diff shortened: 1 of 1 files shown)


d00p:
The quota's meaning is to restrict diskspace usage...not traffic usage...There is not real quota for traffic right now. You can only generate warning-emails to inform the customers about high traffic usage

realbugi:
I understand your point but it's not actually true. See that current version make issues as described in OP. 

When you're calculating quota for ProFTPd service you are using now:
```
panel_customers.traffic*1024 AS bytes_in_avail 
and comparing with 
$current_diskspace['all'] * 1024
```
with is wrong.

This PR solves this issue in correct way I believe (based on traffic).

d00p:
I can only repeat myself....FTP quota (the .ftpquota file) shows the _disk usage_ and **NOT** the _traffic usage_

PR #693, opened 2019-05-27
Fehlermeldung wenn Vendor-Ordner Fehlt

drexlma:
Fehlermeldung wenn Vendor-Ordner Fehlt

### DIFF
--- a/lib/init.php
+++ b/lib/init.php
@@ -16,6 +16,10 @@
  * @package    System
  *
  */
+if(!file_exists(dirname(__DIR__) . '/vendor/autoload.php')){
+    die('Vendor does not exist. Please run "Composer install". For more informationen https://github.com/Froxlor/Froxlor/wiki');
+}
+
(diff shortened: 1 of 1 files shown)


lib/init.php
```diff
@@ -16,6 +16,10 @@
  * @package    System
  *
  */
+if(!file_exists(dirname(__DIR__) . '/vendor/autoload.php')){
+    die('Vendor does not exist. Please run "Composer install". For more informationen https://github.com/Froxlor/Froxlor/wiki');
```
d00p: composer in lowercase letters please...

d00p:
It would be more "froxlor-like" if we would die with a nice hint (html/css) like we do in https://github.com/Froxlor/Froxlor/blob/master/install/lib/class.FroxlorInstall.php#L1089 or https://github.com/Froxlor/Froxlor/blob/master/lib/init.php#L78 - just create a template to display the hint nicely

PR #707, opened 2019-07-28
Update Travis

makuser:
Update Travis to run tests with php7.3 and with mysql 5.6, 5.7 and 8.0 and mariadb 10.3 and 10.4.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,8 +46,9 @@ before_install:
     done;
 
 install:
-  - mysql -h 127.0.0.1 --protocol=TCP -u root -pfr0xl0r.TravisCI -e "CREATE DATABASE IF NOT EXISTS froxlor010"
-  - mysql -h 127.0.0.1 --protocol=TCP -u root -pfr0xl0r.TravisCI -e "GRANT ALL ON froxlor010.* TO 'froxlor010'@'%' IDENTIFIED BY 'fr0xl0r.TravisCI';"
+  - mysql -h 127.0.0.1 --protocol=TCP -u root -pfr0xl0r.TravisCI -e "CREATE DATABASE IF NOT EXISTS froxlor010;"
+  - mysql -h 127.0.0.1 --protocol=TCP -u root -pfr0xl0r.TravisCI -e "CREATE USER 'froxlor010'@'%' IDENTIFIED BY 'fr0xl0r.TravisCI';"
+  - mysql -h 127.0.0.1 --protocol=TCP -u root -pfr0xl0r.TravisCI -e "GRANT ALL ON froxlor010.* TO 'froxlor010'@'%';"
   - mysql -h 127.0.0.1 --protocol=TCP -u root -pfr0xl0r.TravisCI froxlor010 < install/froxlor.sql
 
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ matrix:
     - php: 7.3
       env: "DOCKER_MYSQL_TYPE=mysql DOCKER_MYSQL_VERSION=5.7"
     - php: 7.3
-      env: "DOCKER_MYSQL_TYPE=mysql DOCKER_MYSQL_VERSION=8.0"
+      env: "DOCKER_MYSQL_TYPE=mysql DOCKER_MYSQL_VERSION=8.0 STARTCMD='mysqld --default-authentication-plugin=mysql_native_password'"
     - php: 7.3
       env: "DOCKER_MYSQL_TYPE=mariadb DOCKER_MYSQL_VERSION=10.3"
     - php: 7.3
@@ -29,7 +29,7 @@ addons:
 
 before_install:
   - export MYSQL_DATABASE=froxlor010
-  - docker run -d --name mysql -e MYSQL_ROOT_PASSWORD=fr0xl0r.TravisCI -e MYSQL_DATABASE=$MYSQL_DATABASE -p 3306:3306 $DOCKER_MYSQL_TYPE:$DOCKER_MYSQL_VERSION
+  - docker run -d --name mysql -e MYSQL_ROOT_PASSWORD=fr0xl0r.TravisCI -e MYSQL_DATABASE=$MYSQL_DATABASE -p 3306:3306 $DOCKER_MYSQL_TYPE:$DOCKER_MYSQL_VERSION $STARTCMD
   - sudo apt-get install -y ant
   - >
     export tries=0;
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,14 +12,14 @@ branches:
 
 matrix:
   include:
-#    - php: 7.3
-#      env: "DOCKER_MYSQL_TYPE=mysql DOCKER_MYSQL_VERSION=5.6"
-#    - php: 7.3
-#      env: "DOCKER_MYSQL_TYPE=mysql DOCKER_MYSQL_VERSION=5.7"
-#    - php: 7.3
-#      env: "DOCKER_MYSQL_TYPE=mysql DOCKER_MYSQL_VERSION=8.0"
-#    - php: 7.3
-#      env: "DOCKER_MYSQL_TYPE=mariadb DOCKER_MYSQL_VERSION=10.3"
+    - php: 7.3
+      env: "DOCKER_MYSQL_TYPE=mysql DOCKER_MYSQL_VERSION=5.6"
+    - php: 7.3
+      env: "DOCKER_MYSQL_TYPE=mysql DOCKER_MYSQL_VERSION=5.7"
+    - php: 7.3
+      env: "DOCKER_MYSQL_TYPE=mysql DOCKER_MYSQL_VERSION=8.0"
+    - php: 7.3
+      env: "DOCKER_MYSQL_TYPE=mariadb DOCKER_MYSQL_VERSION=10.3"
     - php: 7.3
       env: "DOCKER_MYSQL_TYPE=mariadb DOCKER_MYSQL_VERSION=10.4"
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,9 @@
 language: php
-dist: trusty
-
+dist: bionic
+services:
+  - docker
 php:
-  - "7.3"
+  - 7.3
 
 branches:
   only:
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -1,8 +1,8 @@
 language: php
-dist: trusty
-
+services:
+  - docker
```
d00p: can you please explain why this is necessary...i'd really like to test as "normal" as possible and not using any docker mysql stuff...

tests/bootstrap.php
```diff
@@ -159,7 +159,7 @@
 Settings::Set('system.documentroot_use_default_value', '1', true);
 Settings::Set('system.hostname', 'dev.froxlor.org', true);
 Settings::Set('system.nameservers', 'dev.froxlor.org', true);
-Settings::Set('system.mysql_access_host', 'localhost,127.0.0.1,2a01:440:1:12:82:149:225:46,82.149.225.46', true);
+Settings::Set('system.mysql_access_host', 'localhost,127.0.0.1,172.17.0.1,2a01:440:1:12:82:149:225:46,82.149.225.46', true);
```
d00p: what's `172.17.0.1` doing in there?

.travis.yml
makuser: Well, that is necessary because you do not only want to test against a single version of MySQL. You don't even want to test against MySQL (the Oracle/SUN one) alone, but also against MariaDB. Spinning up different docker instances with different mysql/mariadb versions allows you exactly that.

> Update Travis to run tests with php7.3 and with mysql 5.6, 5.7 and 8.0 and mariadb 10.3 and 10.4.

tests/bootstrap.php
makuser: Allow the docker host to access the docker instance.

.travis.yml
d00p: well but it acutally does only test using mariadb 10.4 doesn't it?

tests/bootstrap.php
d00p: but you're using `127.0.0.1` in all connections, just don't see the need for this then :) it's more or less just for customer database users - but default setting for most mysql/mariadb servers is they are bound to localhost/127.0.0.1 anyway :)

.travis.yml
makuser: > well but it acutally does only test using mariadb 10.4 doesn't it?

I see you understood the `matrix` object :)
At the moment it's configured for 10.4, because I did not want to spin up 5 jobs per commit while I am testing the changes in the yaml. But when I'm done, I'd reenable the tests.

tests/bootstrap.php
makuser: > but you're using `127.0.0.1` in all connections, just don't see the need for this then :) it's more or less just for customer database users - but default setting for most mysql/mariadb servers is they are bound to localhost/127.0.0.1 anyway :)

I am talking to `127.0.0.1:3306` which forwards packets to `172.17.0.2:3306` where docker is running on.
This option is not what the DST host can be, but what the SRC host can be. And when accessing `172.17.0.2:3306` (DST) the SRC is `172.17.0.1`, so this needs to be added to the allowed hosts.

.travis.yml
d00p: Oh gee, yeah, totally not seen the `matrix` tag - sorry :P nicely done

tests/bootstrap.php
d00p: as long as it works :P it's for testing only anyway :) I never got too deep into this whole docker thing - thanks for your work, it's really appreciated

d00p:
So, let's enable the matrix and it's good to merge, aight?

tests/bootstrap.php
makuser: I had to get into docker for work and I like it even less than before, but in this case (travis-ci matrix) its unfortunately the only way to test it with different database servers :/

makuser: BTW so you can understand why I really had to add the IP `172.17.0.1`: https://travis-ci.com/Froxlor/Froxlor/builds/120991797
This job failed because of:
```
  [phpunit] There was 1 failure:
  [phpunit] 
  [phpunit] 1) MysqlsTest::testCustomerMysqlsAdd
  [phpunit] SQLSTATE[HY000] [1045] Access denied for user 'test1sql1'@'172.17.0.1' (using password: YES)
  [phpunit] 
  [phpunit] /home/travis/build/Froxlor/Froxlor/tests/Mysqls/MysqlsTest.php:45
```

d00p: > I had to get into docker for work and I like it even less than before, but in this case (travis-ci matrix) its unfortunately the only way to test it with different database servers :/

but definetly a very nice feature in this case and really helpful, thanks for that again

makuser:
Alright, so now we should have the system running on Bionic, so 18.04 and testing it with the database server Oracle MySQL 5.6, 5.7 and 8.0 and MariaDB 10.3 and 10.4.

d00p:
Hmm
`SQLSTATE[HY000] [2054] The server requested authentication method unknown to the client`

Do we need something like that: https://forum.froxlor.org/index.php?/topic/13916-setup-proccess-failed-access-denied-for-user-rootlocalhost/&do=findComment&comment=35113

makuser:
Oh maaaaan, Oracle s****! They have removed the shorthand for creating user + permissions in MySQL 8.0. WHY would they do that? Never ever using Oracle anything again...

OSS FTW!

Well, at least now we see that lots of stuff does not actually work properly on Oracle MySQL. Whether that's a good thing or a bad one that needs fixing, well.... Not my problem :tongue:
MariaDB works flawlessly :heart:.

makuser:
So, all the Extrawürste in .travis.yml required for MySQL 8.0 to work seem to be implemented.

The faults seen in MySQL 5.6: https://travis-ci.com/Froxlor/Froxlor/jobs/220516870 (does not support `IF EXISTS`), MySQL 5.7: https://travis-ci.com/Froxlor/Froxlor/jobs/220516871 (https://github.com/Froxlor/Froxlor/blob/master/lib/Froxlor/Database/Database.php#L191 field length, don't forget the int appended at the end) and MySQL 8.0: https://travis-ci.com/Froxlor/Froxlor/jobs/220516872 are in fact problems that need to be solved in Froxlor, eg. DB Manager, not Travis CI.

@d00p Is it ready to merge then, or did you find anything else?

build.xml
```diff
@@ -241,7 +241,7 @@
 
 	<target name="phpunit" unless="phpunit.done" depends="phpunit-prepare"
 		description="Run unit tests with PHPUnit">
-		<exec executable="${phpunit}" resultproperty="result.phpunit"
+		<exec executable="${phpunit}" failonerror="true" resultproperty="result.phpunit"
```
makuser: Just throwing this in there, because that is something that should have been added anyway, because we actually want to fail the entire build when some tests fail, so we know that something is up.

makuser: This job is supposed to fail too, but since the Tests did not fail, the job will not fail either, even though it really should: https://travis-ci.com/Froxlor/Froxlor/jobs/220526962
I assume that's a bug in the test that deletes a user database, which simply lets these exceptions fly.
https://travis-ci.com/Froxlor/Froxlor/jobs/220526963 (MySQL 5.7) failed, which was expected.
https://travis-ci.com/Froxlor/Froxlor/jobs/220526964 (MySQL 8.0) failed too, nice. That's how it should be.

d00p:
> The faults seen in MySQL 5.6: https://travis-ci.com/Froxlor/Froxlor/jobs/220516870 (does not support `IF EXISTS`),

ok, not a big deal I guess

> MySQL 5.7: https://travis-ci.com/Froxlor/Froxlor/jobs/220516871 (https://github.com/Froxlor/Froxlor/blob/master/lib/Froxlor/Database/Database.php#L191 field length, don't forget the int appended at the end) 

I can't seem to follow you on this on, what int appended at the end?

> MySQL 8.0: https://travis-ci.com/Froxlor/Froxlor/jobs/220516872 are in fact problems that need to be solved in Froxlor, eg. DB Manager, not Travis CI.

what's the difference between the error with mysql 5.7 and this one?

makuser:
> > The faults seen in MySQL 5.6: https://travis-ci.com/Froxlor/Froxlor/jobs/220516870 (does not support `IF EXISTS`),
> 
> ok, not a big deal I guess

Yes it is a big deal, as it does not allow Froxlor to do what it was supposed to do (delete a database).

> > MySQL 5.7: https://travis-ci.com/Froxlor/Froxlor/jobs/220516871 (https://github.com/Froxlor/Froxlor/blob/master/lib/Froxlor/Database/Database.php#L191 field length, don't forget the int appended at the end)
> 
> I can't seem to follow you on this on, what int appended at the end?

`web123sql5`
5 in this case.

> > MySQL 8.0: https://travis-ci.com/Froxlor/Froxlor/jobs/220516872 are in fact problems that need to be solved in Froxlor, eg. DB Manager, not Travis CI.
> 
> what's the difference between the error with mysql 5.7 and this one?

None.

But none of these errors have anything to do with this PR anymore.

d00p:
> Yes it is a big deal, as it does not allow Froxlor to do what it was supposed to do (delete a database).

Well I realised people can't read minds - yes it's a big deal - but not for this PR is what I meant, sorry :)

> web123sql5
5 in this case.

Problem is the exception is not thrown at all :)  seems like I did the test without keeping in mind that the test-environment might get updated some day :P I'll take a look at this more deeply later when the kids are sleeping

makuser:
> Well I realised people can't read minds - yes it's a big deal - but not for this PR is what I meant, sorry :)
> 
Aah, that's what you meant. Alright great :)

> > web123sql5
> > 5 in this case.
> 
> Problem is the exception is not thrown at all :) seems like I did the test without keeping in mind that the test-environment might get updated some day :P I'll take a look at this more deeply later when the kids are sleeping

I think the way to calculate the max length is wrong, which is what I was trying to say. If your has 10 chars as max username length and you subtract the prefix (or what you call prefix, but is actually more a suffix) length, which for the default prefix is sql, so it has a length of 3, you have 7 chars left.
Customer username is: makuser, length of 7.

Froxlor will then generate this username: makusersql1, which has a length of 7 + 3 **+ 1 = 11**. After makusersql9 it gets even worse, because makusersql10 is created, which has a length of 7 + 3 **+ 2 = 12**. I did not see the length of the appended incremented integer being subtracted in the code, but maybe I missed that part, as I just quickly skimmed through the whole file.

Do you understand now what I mean?

d00p:
> (or what you call prefix, but is actually more a suffix) 

to be perfectly correct, it's an infix :P

And yes, the calculation may not be perfect, but see \Froxlor\Validate\Validate::validateUsername() for what exactly is being done

makuser:
> And yes, the calculation may not be perfect, but see \Froxlor\Validate\Validate::validateUsername() for what exactly is being done

Ah there it is. Alright, so it works until 9 and fails after that. Correct?

d00p:
> > And yes, the calculation may not be perfect, but see \Froxlor\Validate\Validate::validateUsername() for what exactly is being done
> 
> Ah there it is. Alright, so it works until 9 and fails after that. Correct?

`[a-z][a-z0-9]{0,XYZ}` -> can be anything from k1sql2 to k1sql200 as long as it's at least one character to XYZ characters long

PR #710, opened 2019-08-17
Updating another fastcgi_param SCRIPT_FILENAME

TimoStramann:
Use $request_filename instead of $document_root$fastcgi_script_name as described in: https://www.nginx.com/resources/wiki/start/topics/tutorials/config_pitfalls/#use-request-filename-for-script-filename


d00p:
Does `$fastcgi_script_name` here really have the same value as `$1`? Is this tested when using php-fpm?

TimoStramann:
I have tested php-fpm only.
From what I understand `$1` is the path to the script from the URL so `https://domain.com/folder/file.php` would make `$1 = /folder/file.php` and `$document_root$1` makes it `/var/www/folder/file.php` assuming the document root is /var/www. `$1` is equal to `$fastcgi_script_name` in this case. At least this is the result from my recent test and it's working fine.

PR #714, opened 2019-09-23
Get mailbox size with maildirsize file

Bobselp:
# Description

When quota is enabled, dovecot generates a [maildirsize file](https://wiki2.dovecot.org/Quota/Maildir#Maildirsize_file). This file can be used to get the current Maildir size. This is quicker & easier than using du. When quota is not enabled, the file is missing & du is used as usual.

Especially when quota integration in Dovecot is actively used, du sometimes reports too big values, leading to confusion why it's different from the value shown in mail clients. The big values come from various files that usually do not count towards maildir size, like index files.

## Type of change

# How Has This Been Tested?

Tested on local configuration with various use-cases (file there, file not there, new user mailbox etc.)

**Test Configuration**:

* Distribution: Debian Buster
* Webserver: Apache/2.4.38
* PHP: PHP 7.3.9-1~deb10u1
* Dovecot: 2.3.4.1 (f79e8e7e4)

# Checklist:

### DIFF
--- a/lib/Froxlor/Cron/System/MailboxsizeCron.php
+++ b/lib/Froxlor/Cron/System/MailboxsizeCron.php
@@ -38,16 +38,24 @@ class MailboxsizeCron extends \Froxlor\Cron\FroxlorCron
 			$_maildir = \Froxlor\FileDir::makeCorrectDir($maildir['maildirpath']);
 
 			if (file_exists($_maildir) && is_dir($_maildir)) {
-				// 1. Check if maildirsize exists (usually when quota is enabled)
-				$maildirsize = $_maildir.'maildirsize';
-				if (file_exists($maildirsize)) {
+				$maildirsize = \Froxlor\FileDir::makeCorrectFile($_maildir . '/maildirsize');
+				
+				// When quota is enabled and maildirsize file exists, use that to calculate size
+				if (\Froxlor\Settings::Get('system.mail_quota_enabled') == 1 && file_exists($maildirsize)) {
 					\Froxlor\FroxlorLogger::getInstanceOf()->logAction(\Froxlor\FroxlorLogger::CRON_ACTION, LOG_NOTICE, 'found maildirsize file in ' . $_maildir);
 					$file = file($maildirsize, FILE_IGNORE_NEW_LINES | FILE_SKIP_EMPTY_LINES);
-					if (!empty($file[1])) {
-						$emailusage = floatval(explode(' ', $file[1])[0]);
+					// Remove header
+					array_shift($file);
+					$emailusage = 0;
+					// Sum up all the changes (line 2 -> end)
+					foreach ($file as $line) {
+						$parts = explode(' ', $line);
+						if (!empty($parts[0])) {
+							$emailusage += floatval($parts[0]);
+						}
 					}
 				} else {
-					// 2. If maildirsize file does not exist, compute with du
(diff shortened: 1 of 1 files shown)


lib/Froxlor/Cron/System/MailboxsizeCron.php
```diff
@@ -38,26 +38,37 @@ public static function run()
...
-					'$',
-					'~',
-					'?'
-				));
-				foreach ($back as $backrow) {
-					$emailusage = explode(' ', $backrow);
-				}
-				$emailusage = floatval($emailusage['0']);
+				// 1. Check if maildirsize exists (usually when quota is enabled)
+				$maildirsize = $_maildir.'maildirsize';
```
d00p: Could we change that line to the followig, just to be sure? More froxlor-style xD
```
$maildirsize = \Froxlor\FileDir::makeCorrectFile($_maildir . '/maildirsize');
```

Bobselp:
I fixed a logical error in calculation (the file gets updated every received/send mail, I just looked at the initial value).

I also added a check if quota is enabled. The reason is, with disabled quota (= probably not configured/enabled in dovecot), the maildirsize might be outdated, I'm not sure if dovecot keeps updating it since it's a feature of the quota module.

lib/Froxlor/Cron/System/MailboxsizeCron.php
```diff
@@ -38,16 +38,24 @@ public static function run()
...
+				// When quota is enabled and maildirsize file exists, use that to calculate size
+				if (\Froxlor\Settings::Get('system.mail_quota_enabled') == 1 && file_exists($maildirsize)) {
 					\Froxlor\FroxlorLogger::getInstanceOf()->logAction(\Froxlor\FroxlorLogger::CRON_ACTION, LOG_NOTICE, 'found maildirsize file in ' . $_maildir);
 					$file = file($maildirsize, FILE_IGNORE_NEW_LINES | FILE_SKIP_EMPTY_LINES);
-					if (!empty($file[1])) {
-						$emailusage = floatval(explode(' ', $file[1])[0]);
+					// Remove header
+					array_shift($file);
+					$emailusage = 0;
+					// Sum up all the changes (line 2 -> end)
```
d00p: the dovecot docs dont say how they store the sizes there, only about the header. You say "sum up changes", if the file only contains changes to the size, we won't get the total usage; if I read that correctly from http://www.courier-mta.org/imap/README.maildirquota.html it should be the complete size, right?

Bobselp: My comment is a bit misleading, sorry. Line 1 contains the quota limit, line 2 the start value, line 3-X the changes to that start value.

Each new/removed mail will just add a new line to the file. This saves MTAs from the hassle of recalculating the mailbox size every time a user makes a change. As mentioned on your linked page, after the file size reaches 5120 bytes (due to adding new lines all the time), the MTA regenerates it, so you don't have to walk through thousands of calculations at some point. So the maildirsize always has at least two lines.

This is an example of a maildirsize file after some usage:

```
1073741824S
324923 155
1901 1
1901 1
1901 1
-39790 -20
```

That mailbox (max size: 1GB) starts with 155 mails sized 317KB, receives three new mails with ~2KB each, before I delete 20 mails. At the end, the mailbox has 138 mails with a total size of 284KB

d00p: Got ya, neat :)

PR #745, opened 2019-10-30
Add UnitTests for #679

pquerner:
After #679 is now merged, this adds some unit tests for `Validate::is_ipv6`

### DIFF
--- a/tests/Froxlor/ValidateTest.php
+++ b/tests/Froxlor/ValidateTest.php
@@ -109,13 +109,27 @@ class ValidateTest extends TestCase
         $this->assertEquals("8.8.8.8/128.0.0.0", $result);
     }
 
+    public function testIPv6Checks()
+    {
+        $result = Validate::is_ipv6('1.1.1.1/4');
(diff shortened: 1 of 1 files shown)


d00p:
Ok, and what about replacing the `filter_var` on IPv6 checking with the new `is_ipv6` function?

tests/Froxlor/ValidateTest.php
```diff
@@ -109,13 +109,27 @@ public function testValidateCidrNoationToNetmaskNotationIPv4()
         $this->assertEquals("8.8.8.8/128.0.0.0", $result);
     }
 
+    public function testIPv6Checks()
```
d00p: please stay conform with the naming regarding the rest of the tests, ergo `testValidateIPv6` would be more suitable

pquerner:
1 step at a time. First Unittests, then the world.

tests/Froxlor/ValidateTest.php
pquerner: Changed

PR #763, opened 2019-11-29
SQL error in lib/Froxlor/Api/Commands/Certificates.php

RipClaw2971:
An SQL error occurs when using the SSL certificate search field.


d00p:
One day too late, darn...yes, there is also a `WHERE` clause in the listing, hence an `append` is needed. Looks like 0.10.9 :/

PR #776, opened 2019-12-13
Remove comments rerferring to mua_* variables

crazy4chrissi:
# Description

In the postfix `master.cf` configfiles that Froxlor suggests for some distris, these lines occur:

    #  -o smtpd_client_restrictions=$mua_client_restrictions
    #  -o smtpd_helo_restrictions=$mua_helo_restrictions
    #  -o smtpd_sender_restrictions=$mua_sender_restrictions
    
However, the configfiles do not contain a definition of these mua_* parameters. Thus, if somebody justs uncomments these lines, it will lead to these warnings:

    postconf: warning: /etc/postfix/master.cf: undefined parameter: mua_sender_restrictions
    postconf: warning: /etc/postfix/master.cf: undefined parameter: mua_client_restrictions
    postconf: warning: /etc/postfix/master.cf: undefined parameter: mua_helo_restrictions

Furthermore, the settings in `main.cf` will be overwritten by an empty setting, which can be dangerous as it allows to much.

## Type of change

# How Has This Been Tested?

**Test Configuration**:

* Distribution: Debian Stretch
* Mailserver: Postfix

# Checklist:

Strictly speaking only comments are removed, so no breaking changes possible.

### DIFF
--- a/lib/configfiles/bionic.xml
+++ b/lib/configfiles/bionic.xml
@@ -2372,9 +2372,6 @@ smtp      inet  n       -       y       -       -       smtpd
 #  -o smtpd_tls_security_level=encrypt
 #  -o smtpd_sasl_auth_enable=yes
 #  -o smtpd_reject_unlisted_recipient=no
-#  -o smtpd_client_restrictions=$mua_client_restrictions
-#  -o smtpd_helo_restrictions=$mua_helo_restrictions
-#  -o smtpd_sender_restrictions=$mua_sender_restrictions
 #  -o smtpd_recipient_restrictions=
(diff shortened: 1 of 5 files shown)


d00p:
These are the packages defaults. We try to keep as much default content in the config-files as possible therefore, I don't see the necessity to remove it

crazy4chrissi:
Okay, then instead define the `mua_*` parameters in the `main.cf` ?

d00p:
I repeat: These are the packages defaults. We try to keep as much default content in the config-files as possible

crazy4chrissi:
Okay, probably I should better head over to the package maintainers of the Debian/Ubuntu postfix package.

PR #779, opened 2019-12-25
Add 2FA GA secret output for software TOTP

OnlineArts:
# Description

For 2FA TOTP Software users its use full to know the 2FA secret. Therefore this pull request adds a textual representation of the 2FA secret below the 2FA QR code.

## Type of change

# Checklist:


2fa.php
```diff
@@ -86,5 +86,6 @@
 } elseif ($userinfo['type_2fa'] == '2') {
 	// authenticator 2fa enabled
 	$ga_qrcode = $tfa->getQRCodeImageAsDataUri($userinfo['loginname'], $userinfo['data_2fa']);
+        $ga_secret = &$userinfo['data_2fa'];
```
d00p: why pass that as reference of the variable? Makes absolutely no sense and is not necessary

PR #781, opened 2019-12-31
Use logtype for access_log for IPsAndPorts too

patschi:
# Description

`logtype` can be defined in the settings and should be used consistently across all `access_log` options.

## Type of change

# How Has This Been Tested?

- [X] Configs generated and checked if correct.

# Checklist:


patschi:
I'm not an expert with apache, but apparently the config is fine and the custom LogFormat is used everywhere - I've checked that in  `lib/Froxlor/Cron/Http/Apache.php`.

PR #783, opened 2019-12-31
Ability to add custom config to PHPFPM version

patschi:
# Description

Adds the ability to append custom configuration to PHPFPM pool configuration:
image

Addresses #643

## Type of change
 
- [X] New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

- [X] Tested on Froxlor 0.10.10 installation
- [X] Also tested variables and checked if configs written correctly

# Checklist:

- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] My changes generate no new warnings

PR #785, opened 2019-12-31
Allow using more advanced webserver LogFormat

patschi:
**Important note:** This PR depends on #783 due to DBVERSION number.

# Description

I'm using a more complex LogFormat to log in JSON format, which is then processed further. Unfortunately the current implementation restricts you from using more advanced LogFormats due to two reasons:

### The reasons for this change
1. **The hardcoded quotes**

When using 
```
{"timestamp":"$time_iso8601","remote_addr":"$remote_addr"}
```
...as the LogFormat this will result in...
```
log_format frx_custom "{"timestamp":"$time_iso8601","remote_addr":"$remote_addr"}"
```
So destroying the configuration or require confusing escaping in the whole string. When the string is way longer, it's getting quite unhandy (see example below).

2. Using specific options of LogFormat

For example with nginx you can set the type of escaping - [see docs](https://nginx.org/en/docs/http/ngx_http_log_module.html#log_format) - which cannot be used in the current implementation. This might be required for example when logging in JSON format.

Practical example: When using 
```
escape=json '{"timestamp":"$time_iso8601","remote_addr":"$remote_addr"}'
```
...as the LogFormat, it will result in...
```
log_format frx_custom "escape=json '{"timestamp":"$time_iso8601","remote_addr":"$remote_addr"}'"
```
Which won't work out.

### My usecase
My usecase, as mentioned above, is logging in JSON format. For this I'm using following configuration below:
```
log_format frx_custom escape=json '{"timestamp":"$time_iso8601","remote_addr":"$remote_addr","remote_user":"$remote_user","body_bytes_sent":$body_bytes_sent,"request_time":$request_time,"status":$status,"request":"$request","request_method":"$request_method","request_uri":"$request_uri","uri":"$uri","http_referrer":"$http_referer","x-forwarded-for":"$http_x_forwarded_for","hostname":"$host","server_name":"$server_name","upstream_addr":"$upstream_addr","upstream_status":"$upstream_status","upstream_response_time":"$upstream_response_time","upstream_response_length":"$upstream_response_length","upstream_cache_status":"$upstream_cache_status","ssl":"$https","ssl_protocol":"$ssl_protocol","ssl_cipher":"$ssl_cipher","ssl_session_reused":"$ssl_session_reused","gzip_ratio":"$gzip_ratio","sent_http_content_type":"$sent_http_content_type","http_user_agent":"$http_user_agent"}'
```

My goal was to allow more advanced LogFormats while not breaking any existing installations. So what I basically did is following:
1. Removing the hardcoded quotes from log_format (nginx) / LogFormat (Apache)
2. During update process adding the previously hardcoded quotes in the database (basically migration)
3. Allowing users to set quotes itself, therefore allowing more complex scenarios. Also added hints that wrong configuration might break webserver.

It currently looks like: 
image

## Type of change

# How Has This Been Tested?

- [X] Using my log_format as provided below successfully on my installation with nginx
- [ ] Tested the Apache2 config
- [ ] Tested the migration for older installation

(Latter both options not tested, in all honesty. I don't have apache2 in use)

# Checklist:


install/froxlor.sql
```diff
@@ -703,7 +703,7 @@ opcache.interned_strings_buffer'),
 	('panel', 'customer_hide_options', ''),
 	('panel', 'is_configured', '0'),
 	('panel', 'version', '0.10.10'),
-	('panel', 'db_version', '201912310');
+	('panel', 'db_version', '201912312');
```
d00p: You said this RP depends on #783 but you are not using the source code of it, hence the conflicts now

patschi: Yeah, I'm honestly not sure how you/in general handle multiple features with DB changes at once without conflicting with each other. I've each time used the `master` branch for each PR (without my other changes) - this way it would have been easier to merge PRs with less changes on each own, without depending on each other code-wise.

I've solved the conflicts.

d00p: Yes basically that's correct of course. But if you depend on other features you would use the the other features code-base as "master" - anyway, these "conflicts" are not unfixable but it's of course nicer if there is less code-trouble :)

patschi:
I believe you were referring to PR#786 for awstats instead. I've pushed the changes there as suggested.

d00p:
> I believe you were referring to PR #786 for awstats instead. I've pushed the changes there as suggested.

Well, it's for both PR's as they belong together anyway :)

PR #786, opened 2019-12-31
Allow configuring awstats LogFormat

patschi:
**Important note**: This PR depends on #785 due to DBVERSION number.

# Description

When using custom LogFormat on webservers, awstats is usually not able to correctly parse the logs and therefore also the traffic calculation fails. With this PR it's possible to change the LogFormat of awstats, which is then used when awstats is called.

Screenshot of the new setting in the UI:
image

## Type of change

# How Has This Been Tested?

- [X] Tested if default value 1 works
- [X] Tested longer and more complex LogFormat values
- [X] Tested if awstats models are written correctly for the domains

# Checklist:


install/froxlor.sql
```diff
@@ -703,7 +703,7 @@ opcache.interned_strings_buffer'),
 	('panel', 'customer_hide_options', ''),
 	('panel', 'is_configured', '0'),
 	('panel', 'version', '0.10.10'),
-	('panel', 'db_version', '201912310');
+	('panel', 'db_version', '201912313');
```
d00p: Missing the new setting in `install/froxlor.sql` - new installations wont be able to save the setting

patschi: I was pretty sure I did not forget something this time. Well... Thanks for your patience with me! :)

lng/english.lng.php
```diff
@@ -2092,3 +2092,6 @@
 $lng['serversettings']['phpfpm_settings']['restart_note'] = 'Attention: The config won\'t be checked for any errors. If it contains errors, PHP-FPM might not start again!';
 $lng['serversettings']['phpfpm_settings']['custom_config']['title'] = 'Custom configuration';
 $lng['serversettings']['phpfpm_settings']['custom_config']['description'] = 'Add custom configuration to each PHP-FPM version instance, for example <i>pm.status_path = /status</i> for monitoring. Variables below can be used here. ' . ' <strong>' . $lng['serversettings']['phpfpm_settings']['restart_note'] . '</strong>';
+
+$lng['serversettings']['awstats']['logformat']['title'] = 'LogFormat setting';
+$lng['serversettings']['awstats']['logformat']['description'] = 'If you use customized logformat for your webserver, you need change the awstats LogFormat too.<br/>Default is 1. For more information check documentation <a target="_blank" href="https://awstats.sourceforge.io/docs/awstats_config.html#LogFormat">here</a>. Might also require changes in the awstats config.';
```
d00p: why would it require changes to the awstats configuration? There is a placeholder which is replaced by froxlor's traffic cron. Only time the config needs adjustment is when updating...maybe we should add a hint for this in the preconfig (I can do that after merging)

patschi: Then I apparently understood your comment in the other PR wrong. Sorry! I just removed it from the language files and I'm now going over to fix the conflicts.

Beside that: You could also use "Merge & Squash" (or however the option is called exactly) so that it doesn't mess up your clean Git history in the main branch and keeps it clean.

patschi:
I have to mention it... Thanks a lot for your patience with me and your time, @d00p!

actions/admin/settings/140.statistics.php
```diff
@@ -78,6 +79,14 @@
...
 					'default' => '/usr/share/awstats/icon/',
 					'save_method' => 'storeSettingField'
+				),
+				'system_awstats_logformat' => array(
+					'label' => $lng['serversettings']['awstats']['logformat'],
+					'settinggroup' => 'system',
+					'varname' => 'awstats_logformat',
+					'type' => 'string',
+					'default' => '1',
+					'save_method' => 'storeSettingField'
```
d00p: would also add the visible field here
```php
'visible' => \Froxlor\Settings::Get('system.awstats_enabled') == 1
```

PR #795, opened 2020-01-23
Fixed the case that the spf record is not inserted with it's quotes, …

lpedretti:
…and so the condition fails and 2 spf records are inserted in the domain

# Description
Whe adding manual dns records, specifically an spf record, the ["content"] key of the array contains the spf record without the quotes, and so the condition fails, leaving the default spf record alongside the newly created one

## Type of change

# How Has This Been Tested?

Added a custom spf record, and checked the default record was removed
Deleted the custom spf record, and checked that the default record was there again

**Test Configuration**:

* Distribution: Ubuntu 18.04
* Webserver: Apache 2.4
* PHP: 7.2 fpm
* DNS daemon: PowerDNS

### DIFF
--- a/lib/Froxlor/Dns/Dns.php
+++ b/lib/Froxlor/Dns/Dns.php
@@ -160,7 +160,7 @@ class Dns
 				// unset special CAA required-entry
 				unset($required_entries[$entry['type']][md5("@CAA@")]);
 			}
-			if (Settings::Get('spf.use_spf') == '1' && $entry['type'] == 'TXT' && $entry['record'] == '@' && strtolower(substr($entry['content'], 0, 7)) == '"v=spf1') {
+			if (Settings::Get('spf.use_spf') == '1' && $entry['type'] == 'TXT' && $entry['record'] == '@' && (strtolower(substr($entry['content'], 0, 7)) == '"v=spf1' || strtolower(substr($entry['content'], 0, 6)) == 'v=spf1') ) {
 				// unset special spf required-entry
 				unset($required_entries[$entry['type']][md5("@SPF@")]);
 			}


lib/Froxlor/Cron/Http/Apache.php
```diff
@@ -168,7 +168,7 @@ public function createIpPort()
 				$without_vhost = $this->virtualhosts_data[$vhosts_filename];
 				$close_vhost = true;
 
-				$this->virtualhosts_data[$vhosts_filename] .= '<VirtualHost ' . $ipport . '>' . "\n";
+				$this->virtualhosts_data[$vhosts_filename] .= '<VirtualHost *:' . $row_ipsandports['port'] . '>' . "\n";
```
d00p: Ehm, no...this has nothing to do with quotes in DNS/SPF entries...

lib/Froxlor/Cron/Http/Apache.php
```diff
@@ -907,7 +907,8 @@ protected function getVhostContent($domain, $ssl_vhost = false)
 			if (filter_var($domain['ip'], FILTER_VALIDATE_IP, FILTER_FLAG_IPV6)) {
 				$ipport = '[' . $domain['ip'] . ']:' . $domain['port'] . ' ';
 			} else {
-				$ipport = $domain['ip'] . ':' . $domain['port'] . ' ';
+				//$ipport = $domain['ip'] . ':' . $domain['port'] . ' ';
+				$ipport = '*:' . $domain['port'] . ' ';
```
d00p: Same here...just no

lib/Froxlor/Cron/Http/Apache.php
lpedretti: This change wasn't supposed to go here, i did the pull request before doing this commit to my clone of the repository, i'm looking on how do i edit the pull request to remove it.

lib/Froxlor/Cron/Http/Apache.php
lpedretti: Yep, see the other wrong change comment, i'm checking on how to edit the pull request

lib/Froxlor/Cron/Http/Apache.php
lpedretti: Ok, i was able to change the pull request and remove these changes, as they were personal to my installation and didn't want to go into it.

PR #800, opened 2020-02-04
Logging --force parameter

drexlma:
Logging --force parameter


lib/Froxlor/Cron/MasterCron.php
```diff
@@ -72,6 +72,8 @@ public static function run()
 					\Froxlor\System\Cronjob::inserttask('99');
 					array_push($jobs_to_run, 'tasks');
 					define('CRON_IS_FORCED', 1);
+					self::$cronlog->logAction(\Froxlor\FroxlorLogger::CRON_ACTION, LOG_NOTICE, 'active --force');
```
d00p: 1) I'd rather use something like "Cronjob was forced via parameter"
2) Why not do the same for `--debug` and `--no-fork`?

d00p:
closed due to inactivity

PR #819, opened 2020-03-21
fixed parsing due to changes in dovecots default mail_log_prefix

HBerni:
Fixed parsing due to changes in dovecots default mail_log_prefix since version 2.3
Regex supports old and new default format now
See https://wiki2.dovecot.org/Upgrading/2.3 for changes in mail_log_prefix default settings

# Description

Dovecot traffic was not counted anymore, since dovecot changed the default log prefix

## Type of change

- [ x] Bug fix (non-breaking change which fixes an issue)


d00p:
Thank you very much

Issue #820, opened 2020-03-29
Renewal of Let's Encrypt ECC / ECDSA certificates is skipped


chrschn:
# Summary

Renewal of Let's Encrypt certificates fails if a  ECC / ECDSA key length is specified in the admin settings. The ACME v2 script requires the flag `--ecc` to renew ECC certificates, even if a the key length is prefixed with `ec-`.

# System information
* Froxlor version: 0.10.15
* Web server: apache2
* DNS server: Bind
* POP/IMAP server: Dovecot
* SMTP server: postfix
* FTP server: proftpd
* OS/Version: Debian GNU/Linux 10 (buster)

# Steps to reproduce

1. In the Froxlor admin config, enable Let's Encrypt and set a non-zero key length for ECC/ECDSA.
2. Enable Let's Encrypt for a domain and wait for a certificate to be issued.
3. Wait until the certificate expires.
4. Run the cronjob manually `/usr/bin/php -q /var/www/froxlor/scripts/froxlor_master_cronjob.php --letsencrypt --debug`

# Expected behavior

1. Certificates are properly renewed.

# Actual behavior

1. The `acme.sh` script throws an error message:
`The domain 'example.com' seems to have a ECC cert already, please add '--ecc' parameter if you want to use that cert.`

# Log files/log entries
Here is a sample log from the cronjob run with `--debug` set and also increased verbosity for `acme.sh` (actual domain name replaced):
<pre>
[information] Requesting/renewing Let's Encrypt certificates
[information] Updating certificate for example.com
[information] Adding SAN entry: example.com
[information] Adding SAN entry: www.example.com
[information] Checking for LetsEncrypt client upgrades before renewing certificates:
[So 29. Mar 21:46:04 CEST 2020] Already uptodate!
[So 29. Mar 21:46:04 CEST 2020] Upgrade success!
[So 29. Mar 21:46:04 CEST 2020] Removing cron job
[So 29. Mar 21:46:04 CEST 2020] Lets find script dir.
[So 29. Mar 21:46:04 CEST 2020] _SCRIPT_='/root/.acme.sh/acme.sh'
[So 29. Mar 21:46:04 CEST 2020] _script='/root/.acme.sh/acme.sh'
[So 29. Mar 21:46:04 CEST 2020] _script_home='/root/.acme.sh'
[So 29. Mar 21:46:04 CEST 2020] Using config home:/root/.acme.sh
[...]

PR #821, opened 2020-03-29
Fix renewal of ECC/ECDSA certificates.

chrschn:
# Description

The ACME v2 implementation uses separate directoies for ECC and on-ECC
certificates. The renew command for a domain checks if an ECC directory
exists (having a "_ecc" suffix) and refuses the command unless the
"--ecc" flag was specified.

Confusingly, this flag is only required to *renew* an ECC certificate,
but not to issue it.

Fixes https://github.com/Froxlor/Froxlor/issues/820.

## Type of change

# How Has This Been Tested?

- [x] I ran the cronjob which successfully renewed my previously expired ECC certificates.

**Test Configuration**:

* Distribution: Debian GNU/Linux 10 (buster)
* Web server: apache2
* ACME: 2.8.6

# Checklist:

### DIFF
--- a/lib/Froxlor/Cron/Http/LetsEncrypt/AcmeSh.php
+++ b/lib/Froxlor/Cron/Http/LetsEncrypt/AcmeSh.php
@@ -327,6 +327,9 @@ class AcmeSh extends \Froxlor\Cron\FroxlorCron
 			}
 			if (Settings::Get('system.leecc') > 0) {
 				$acmesh_cmd .= " --keylength ec-" . Settings::Get('system.leecc');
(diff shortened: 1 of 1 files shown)


chrschn:
I'm not sure how a change like this should be tested. With this change applied, the `--letsencrypt` cronjob successfully renewed all my ECC certificates, which previously did not work, so at least we got that. I check the [documentation](https://github.com/acmesh-official/acme.sh/blame/master/acme.sh#L6263) of the `--ecc` flag and made sure not to set it when issuing a new certificate, but it is required for renewing and revoking.

PR #838, opened 2020-05-17
Sensible default for domain alias

negrusti:
I believe in 99% of cases www alias is what people want. At the moment the default selection is wildcard, if overlooked by admin it often creates a problem with Letsencrypt certificate issuance.

### DIFF
--- a/admin_domains.php
+++ b/admin_domains.php
@@ -291,7 +291,7 @@ if ($page == 'domains' || $page == 'overview') {
 			// create serveralias options
 			$serveraliasoptions = "";
 			$serveraliasoptions .= \Froxlor\UI\HTML::makeoption($lng['domains']['serveraliasoption_wildcard'], '0', '0', true, true);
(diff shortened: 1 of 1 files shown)


d00p:
In terms of let's encrypt that might be totally true yes. On the other hand, people expect their whole domain to be available (hence the default on wildcard) and not just domain.tld and www.domain.tld.

Another approach would be checking via JavaScript whether let's encrypt was enabled and inform accordingly if alias is wildcard before submitting the form.

negrusti:
@d00p do you think it's possible to create a poll on Froxlor forum to confirm what is the most common use case?

d00p:
Easiest way would be to add a setting what the default for new domains should be ;)

negrusti:
> Easiest way would be to add a setting what the default for new domains should be ;)

That would be great, but that's beyond my understanding of Froxlor internals. I was planning to make more PRs for various options, based on the experience I had with various clients navigating Froxlor UI.

d00p:
A setting for a default value for new domains will be added. as this PR is not even functional (only set the default to '1' to one of the three select-options) it will be closed

d00p:
refs 61a50cc

PR #839, opened 2020-05-17
TLSv1 is deprecated

negrusti:
TLSv1 is deprecated, I believe it should not be enabled for new domains by default. Will be great to enable TLSv1.3 by default but it might be not available on the older distributions.


d00p:
Mozilla's ssl-config for example suggests TLSv1.1 for intermediate compatibility, I think we should do the same

negrusti:
TLSv1.1 is deprecated as well  
https://tools.ietf.org/id/draft-ietf-tls-oldversions-deprecate-02.html  
Admins can enable these if necessary; it's just about a sane default value.

d00p:
Pull requests are always welcome ;)

negrusti:
@d00p I believe my original PR makes the most sense unless you insist on TLSv1.1

PR #841, opened 2020-05-23
Don't suggest php5 during update as default

rubo77:
# Description

During update there is a default suggestion of php5, which is not needed, usually `/etc/alternatives/php` points to the right installed php version.

## Type of change

# How Has This Been Tested?

untested


rubo77:
> Pretty old change for people with < 0.9.32-rc2

I did this when updating from syscp on Debian jessie: I first installed froxlor-0.9.5 then updated to froxlor-0.9.28 (which was the last, I found running without php7)

thx for the merge!

PR #842, opened 2020-05-23
Update RHEL / CentOS configs and set defaults for redhat-based systems

daxara:
# Description

This pull request updates RHEL7 config and adds an RHEL8 config.

Fixes # (issue)

## Type of change

# How Has This Been Tested?

I've ran tests on centos 7 and centos 8 systems (clean installs)

**Test Configuration**:

* Distribution: CentOS 7 + CentOS 8
* Webserver: Apache 2.4
* PHP: 7.2

# Checklist:

- [-] I have made corresponding changes to the documentation
- [-] I have added tests that prove my fix is effective or that my feature works
- [-] New and existing unit tests pass locally with my changes


install/lib/class.FroxlorInstall.php
```diff
@@ -487,6 +487,14 @@ private function _doSettings(&$db)
 		if ($this->_data['webserver'] == "apache24") {
 			$this->_updateSetting($upd_stmt, 'apache2', 'system', 'webserver');
 			$this->_updateSetting($upd_stmt, '1', 'system', 'apache24');
+			
+			// on redhat systems set different defaults
+			if(file_exists('/etc/redhat-release')) {
```
d00p, line 492: too specific, also these are all __settings__ you can adjust, unnecessary here

install/lib/class.FroxlorInstall.php
```diff
@@ -504,6 +512,12 @@ private function _doSettings(&$db)
 			$this->_updateSetting($upd_stmt, 'error', 'system', 'errorlog_level');
 		}
 
+		// on redhat systems set different defaults
+		if(file_exists('/etc/redhat-release')) {
```
d00p, line 516: same as above

lib/configfiles/rhel7.xml
```diff
@@ -61,6 +61,20 @@ Alias "/.well-known/acme-challenge" "{{settings.system.letsencryptchallengepath}
 					<command><![CDATA[systemctl reload-or-restart httpd.service]]></command>
 				</daemon>
 			</service>
+			<!--DNS -->
```
d00p, line 64: Has this been tested to work correctly?

lib/configfiles/rhel8.xml
```diff
@@ -0,0 +1,2448 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
d00p, line 1: any specific changes from 7 to 8?

daxara: none currently,
however RHEL8 is running postfix 3.3 and RHEL7 is running 2.10
By default postfix 3.3 runs in backwards compatibility mode and i have tested it to work with that.
if i get around to it i'll see if the postfix 3.3 config can be optimized and thus changing it for rhel 8.

lib/configfiles/rhel7.xml
daxara: yes.

install/lib/class.FroxlorInstall.php
daxara: Its as unspecific as i could make it (it will work on all the redhat-based distro's).
It would also give an works-out-of-the-box experience similar to that given to debian/ubuntu distro's.

install/lib/class.FroxlorInstall.php
daxara: same answer as above

daxara: see new commit

install/lib/class.FroxlorInstall.php
daxara: see new commit

daxara:
i'll work on an new pull request.

PR #849, opened 2020-05-29
Detect distribution

daxara:
# Description

Add distribution detection and setting defaults if those are given in the corresponding xml file.
it reads /etc/os-release which is a LSB default.
i've looked up the values for supported debian and ubuntu but thave not tested on those distributions.

Fixes # (issue)

## Type of change

# How Has This Been Tested?
- [x] Tested a clean install on CentOS machines with and without the defaults section in the distributions XML file

* Distribution: CentOS 7+ 8

# Checklist:

- [x] I have made corresponding changes to the documentation
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes


install/lib/class.FroxlorInstall.php
```diff
@@ -1273,6 +1298,48 @@ private function _guessWebserver()
 		}
 	}
 
+
+	/**
+	 * get/guess linux distribution
+	 */
+	private function _guessDistribution()
```
d00p, line 1305: too hardcoded... why not check for supported distros (from xml, see https://github.com/Froxlor/Froxlor/blob/master/admin_configfiles.php#L131) and check $os_version data against that to see if it's a match?

install/lib/class.FroxlorInstall.php
```diff
@@ -1273,6 +1298,48 @@ private function _guessWebserver()
...
+	{
+		// post
+		if (! empty($_POST['distribution'])) {
+			$this->_data['distribution'] = $_POST['distribution'];
+		} else {
+			$os_version = parse_ini_file('/etc/os-release', false);
+			if ($os_version['ID'] == 'debian') {
```
d00p, line 1312: open if, no content, elseif check for the same statement...

install/lib/class.FroxlorInstall.php
```diff
@@ -1273,6 +1298,48 @@ private function _guessWebserver()
...
+
+			} elseif ($os_version['ID'] == 'debian') {
+                                if(explode('.',$os_version['VERSION_ID'])[0] == '8') {
```
d00p, line 1315: why explode() everytime in the if-elseif-statements if you can get the version using `$version = explode('.',$os_version['VERSION_ID'])[0];` prior of that and then always using $version

install/lng/english.lng.php
```diff
@@ -63,6 +63,15 @@
 $lng['install']['admin_pass2'] = 'Administrator-Password (confirm)';
 $lng['install']['activate_newsfeed'] = 'Enable the official newsfeed<br><small>(https://inside.froxlor.org/news/)</small>';
 $lng['install']['serversettings'] = 'Server settings';
+$lng['install']['distribution'] = 'Distribution';
+$lng['install']['jessie'] = 'Debian Jessie';
```
d00p, line 67: too hardcoded, you can get the distro names from the xmls

install/lib/class.FroxlorInstall.php
```diff
@@ -840,6 +850,20 @@ private function _showDataForm()
 		 */
 		$section = $this->_lng['install']['serversettings'];
 		eval("\$formdata .= \"" . $this->_getTemplate("datasection") . "\";");
+		// distribution
+		if (! empty($_POST['installstep']) && $this->_data['distribution'] == '') {
+			$diststyle = 'color:red;';
+		} else {
+			$diststyle = '';
+		}
+		$formdata .= $this->_GETsectionItemCheckbox('distribution', 'jessie', ($this->_data['distribution'] == 'jessie'), $diststyle);
```
d00p, line 859: too hardcoded, please read in xml files which distribution is supported, see https://github.com/Froxlor/Froxlor/blob/master/admin_configfiles.php#L131

lib/configfiles/rhel7.xml
```diff
@@ -2,6 +2,38 @@
 <froxlor>
 	<distribution name="RHEL" version="7"
 		defaulteditor="/usr/bin/nano">
+      <defaults>
```
d00p, line 5: that's more or less what I was thinking as distro-dependant setting-defaults. But why `name` and `category`? The settings are in `settinggroup`, `varname` and `value` - we should keep that equal to the rest of the system.

lib/configfiles/rhel7.xml
```diff
@@ -2,6 +2,38 @@
...
+        <propery>
```
d00p, line 6: guess you mean "property". Why not just call it "setting", because that's what it is.

lib/configfiles/rhel7.xml
d00p, line 5: What about defaults for the other distros? :)

install/lib/class.FroxlorInstall.php
```diff
@@ -504,6 +505,15 @@ private function _doSettings(&$db)
 			$this->_updateSetting($upd_stmt, 'error', 'system', 'errorlog_level');
 		}
 
+		if (file_exists(dirname(__DIR__).'/../lib/configfiles/'.$this->_data['distribution'].'.xml')) {
```
d00p, line 508: Why not do it the way we do it in admin_configfiles? See https://github.com/Froxlor/Froxlor/blob/master/admin_configfiles.php#L98

daxara: will do

lib/configfiles/rhel7.xml
daxara: i've only set defaults for RHEL systems as those are the onces i know for certain.
if someone provides defaults for other systems those will be taken into account.

lib/configfiles/rhel7.xml
daxara: yes, typo.
i can rename it, no problem

lib/configfiles/rhel7.xml
daxara: agreed

install/lib/class.FroxlorInstall.php
daxara: ok

install/lng/english.lng.php
daxara: ok

install/lib/class.FroxlorInstall.php
daxara: sure

install/lib/class.FroxlorInstall.php
daxara: sure

install/lib/class.FroxlorInstall.php
daxara: oops

install/lng/german.lng.php
```diff
@@ -63,6 +63,7 @@
 $lng['install']['admin_pass2'] = 'Administrator-Passwort (Bestätigung)';
 $lng['install']['activate_newsfeed'] = 'Aktiviere das offizielle Newsfeed<br><small>(https://inside.froxlor.org/news/)</small>';
 $lng['install']['serversettings'] = 'Servereinstellungen';
+$lng['install']['distribution'] = 'Verteilung';
```
d00p, line 66: 'Verteilung' ist the direct translation of distribution, you can safely use "Distribution" is german too

install/lib/class.FroxlorInstall.php
d00p: Waiting on this

install/lng/german.lng.php
daxara: ah ok.

PR #851, opened 2020-05-31
rename rhel7 to centos7 for consistency

daxara:
# Description

Renames rhel7.xml to centos7.xml for consistency with centos8 config.
also Rhel7 config has CentOS as name in its contents.


d00p:
haha, was about to commit that myself and just that minute your PR pops up :)

PR #852, opened 2020-05-31
Centos8 config updated

daxara:
# Description
This updates the centos8 config with the following changes:
- use DNF instead of YUM (yum is symlinked to dnf in centos8)
- use 'dnf -y install' instead of 'dnf install' to eliminate unnecessary user interaction
- proftpd requires epel repository to be enabled first
- missing create directory for http passwd files
- awstats needs to be installed and requires CentOS PowerTools repository
- awstats cron file is placed in a different directory

# How Has This Been Tested?
- [x] tested installation on clean installed centos 8 machine

# Checklist:

- [X] I have made corresponding changes to the documentation

### DIFF
--- a/lib/configfiles/centos8.xml
+++ b/lib/configfiles/centos8.xml
@@ -64,6 +64,7 @@
 						</command>
 						<command><![CDATA[chown root:0 {{settings.system.apacheconf_diroptions}}]]></command>
 						<command><![CDATA[chmod 0600 {{settings.system.apacheconf_diroptions}}]]></command>
+            <command><![CDATA[mkdir -p {{settings.system.apacheconf_htpasswddir}}]]></command>
 						<command><![CDATA[mkdir -p {{settings.system.documentroot_prefix}}]]></command>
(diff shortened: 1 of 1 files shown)


d00p:
> use 'dnf -y install' instead of 'dnf install' to eliminate unnecessary user interaction

We do the same on apt based distros, it's intended so the admin can review

lib/configfiles/centos8.xml
```diff
@@ -97,7 +98,7 @@ Alias "/.well-known/acme-challenge" "{{settings.system.letsencryptchallengepath}
 			<service type="dns" title="{{lng.admin.configfiles.dns}}">
 				<!--Bind9 -->
 				<daemon name="bind" title="Bind9 nameserver" default="true">
-					<install><![CDATA[yum install bind]]></install>
+					<install><![CDATA[dnf -y install bind]]></install>
```
d00p, line 101: as said, please leave the `-y` out of the command as we have it for apt-based distro's too

daxara: ah ok, will undo

PR #855, opened 2020-05-31
Auto detection in CLI & checks

daxara:
# Description
Adds distribution auto detection to CLI interface

# Checklist:

- [X] I have made corresponding changes to the documentation

### DIFF
--- a/lib/Froxlor/Cli/Action/ConfigServicesAction.php
+++ b/lib/Froxlor/Cli/Action/ConfigServicesAction.php
@@ -92,8 +92,9 @@ class ConfigServicesAction extends \Froxlor\Cli\Action
 		//read os-release
 		if(file_exists('/etc/os-release')) {
 			$os_dist = parse_ini_file('/etc/os-release', false);
-			if(is_array($os_dist) && array_key_exists('ID', $os_dist) && array_key_exists('VERSION_ID', $os_dist))
-			$os_version = explode('.',$os_dist['VERSION_ID'])[0];
+			if(is_array($os_dist) && array_key_exists('ID', $os_dist) && array_key_exists('VERSION_ID', $os_dist)) {
+				$os_version = explode('.',$os_dist['VERSION_ID'])[0];
+			}
 		}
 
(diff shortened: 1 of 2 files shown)


lib/Froxlor/Cli/Action/ConfigServicesAction.php
```diff
@@ -83,6 +83,12 @@ private function createConfig()
 		$distros = glob($config_dir . '*.xml');
 		// tmp array
 		$distributions_select_data = array();
+
+		//read os-release
+		$os_dist = parse_ini_file('/etc/os-release', false);
+		$os_version = explode('.',$os_dist['VERSION_ID'])[0];
```
d00p, line 89: you should at least validate that `$os_dist` is an array and that `['VERSION_ID']` exists...pretty ugly if (why so ever) `/etc/os-release` does not exist

lib/Froxlor/Cli/Action/ConfigServicesAction.php
```diff
@@ -83,6 +83,12 @@ private function createConfig()
...
+		$os_default = "";
```
d00p, line 90: default/fallback should stay "buster" (most commonly used) in case no default is found later on (as said, just fallback, should not happen)

daxara: sure

lib/Froxlor/Cli/Action/ConfigServicesAction.php
daxara: errh.. yeah

daxara: /etc/os-relase is a LSB file,
this is required when enabling the froxlor repository's.
the file should always be there.
nevertheless i added a check for it.

d00p: I know...still, froxlor can basically be installed and used on every system that runs a decent webserver, php and mysql/mariadb database....so it's just in case :)

PR #900, opened 2020-11-23
Protect only private keys and leave certificates world readable.

heptalium:
# Description

Until now all files in Froxlor's SSL directory are only readable by root. While this is required for the .key files which contain the private keys it isn't necessary for the certificates (and chains) which only contain public keys.
Having the certificates world readable allows external programs executed by non-privileged users to scan the certificates. So it would be possible for a monitoring system to check whether there are expired certificates.
As the certificates are presented to all website visitors there is no reason to protect them.

This change executes chmod 600 only for ssl_key_files and leaves all other files mode 644 (or whatever is requested by umask on global level).

## Type of change

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

- [x] Applied the change, requested config rebuild and checked the content of SSL directory and website functionality.

**Test Configuration**:

* Distribution: Debian 10
* Webserver: Apache 2.4.38
* PHP: 7.3

# Checklist:

### DIFF
--- a/lib/Froxlor/Cron/Http/DomainSSL.php
+++ b/lib/Froxlor/Cron/Http/DomainSSL.php
@@ -105,7 +105,9 @@ class DomainSSL
 					$_fh = fopen($filename, 'w');
 					fwrite($_fh, $dom_certs[$type]);
 					fclose($_fh);
-					chmod($filename, 0600);
+					if ($type == 'ssl_key_file') {
+						chmod($filename, 0600);
+					}
 				}
 			}
 			// override corresponding array values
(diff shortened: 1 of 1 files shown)


lib/Froxlor/Cron/Http/DomainSSL.php
```diff
@@ -105,7 +105,9 @@ public function setDomainSSLFilesArray(array &$domain = null)
 					$_fh = fopen($filename, 'w');
 					fwrite($_fh, $dom_certs[$type]);
 					fclose($_fh);
-					chmod($filename, 0600);
```
d00p, line 108: depending on the umask/defaults, it might be a good idea to also explicitly set the 0644 for any other type

heptalium: OK, maybe it is better to have a defined state than use the default.

d00p (approved):

Issue #904, opened 2021-01-08
Email validation bug


bloodhunterd:
_Froxlor version_: **master**/**0.10.23.1**

I encountered an error on the panel setting page when I try to change the sender email address.

![image](https://user-images.githubusercontent.com/43681086/104047286-b456a200-51e1-11eb-8dc1-91da35fdbc1c.png)

The reason is that the **validateEmail** function uses **filter_var** that returns the filtered data on success. So in this case the email address as string.

![image](https://user-images.githubusercontent.com/43681086/104048566-b6b9fb80-51e3-11eb-8f89-daf5a8bddcf9.png)

The validateEmail function is called via detours by the **validateFormField** function.

![image](https://user-images.githubusercontent.com/43681086/104048946-5e372e00-51e4-11eb-9a20-73d747c23c9e.png)

The **standard_error** will be called, because the email address string isn't _identical_ to the required boolean **TRUE**.

I fixed this error already in my fork and I can provide a PR if you want.

d00p:
Weird, return-value if `validateEmail()` hasnt changed since forever.

> I fixed this error already in my fork and I can provide a PR if you want.

Sure, thanks

PR #906, opened 2021-01-12
Fix issue #904 Email validation bug

bloodhunterd:
# Description

The if statement cares about that the **validateEmail** function always returns a boolean value.

Fixes # 904

## Type of change

# How Has This Been Tested?

Debugged the code and tested directly with debug statements.

**Test Configuration**:

* Distribution: Debian
* Webserver: NGINX
* PHP: 7.4

### DIFF
--- a/lib/Froxlor/Validate/Validate.php
+++ b/lib/Froxlor/Validate/Validate.php
@@ -22,10 +22,10 @@ class Validate
 	 * @param
 	 *        	string language id for the error
 	 * @return string the clean string
-	 *        
+	 *
 	 *         If the default pattern is used and the string does not match, we try to replace the
 	 *         'bad' values and log the action.
-	 *        
+	 *
 	 */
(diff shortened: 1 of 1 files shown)


d00p (changes requested):
Travis fails as you just changed the return type of that function ... it is supposed to return the email address if valid, usages like this occur:

 `$email = \Froxlor\Validate\Validate::validateEmail($_POST['loginemail'], 'email');`

d00p:
refs #904

bloodhunterd:
Ah I see...
So my proposal would be to seek for all of the occurrences of this construct in the code and separeted the validation check from the return value like this:
```
$email = \Froxlor\Validate\Validate::validateEmail($_POST['loginemail'], 'email') ? $_POST['loginemail'] : false;
```
But that would mean that the test also have to be adjusted. Are you fine with this?

d00p:
No, the problem is within the settings-saving, not the validation or its return or usage elesewhere

PR #919, opened 2021-02-21
dns: make mail address of SOA records configurable

nachtgeist:
Quoting the default regex delimiter is required for the password
complexity check to work if '/' had been specified as special character
in Froxlor's account settings.

### DIFF
--- a/install/froxlor.sql
+++ b/install/froxlor.sql
@@ -674,6 +674,7 @@ opcache.interned_strings_buffer'),
 	('system', 'apply_phpconfigs_default', '1'),
 	('system', 'hide_incompatible_settings', '0'),
 	('system', 'include_default_vhostconf', '0'),
+	('system', 'soaemail', ''),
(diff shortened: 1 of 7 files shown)


install/froxlor.sql
```diff
@@ -709,7 +709,7 @@ opcache.interned_strings_buffer'),
 	('panel', 'customer_hide_options', ''),
 	('panel', 'is_configured', '0'),
 	('panel', 'version', '0.10.24'),
-	('panel', 'db_version', '202101200');
+	('panel', 'db_version', '202102200');
```
d00p, line 712: You also need to add the new setting here for new installations:

d00p (changes requested):
Thanks for your work, nice addition

install/froxlor.sql
nachtgeist: silly me…fixed

d00p (approved):

PR #929, opened 2021-03-21
pdns: default zone type should be master

Bobselp:
# Description

By default, pdns.conf is created with "master = yes" and pre-sets for XFR enabled. For it to properly work, the zone type of new domains should be MASTER, not NATIVE.

## Type of change


d00p:
I remember something with that being a problem back in the day - which pdns version are you using?

d00p:
no feedback for 3 months

PR #933, opened 2021-04-08
Update Extrausers.php

tokijici:
Migrating from mysql 5.7 to mysql 8 something in the default sorting order has probably changed, default ownership of the customer web folders have switched to ftp users instead of the default username, for example running the query:
SELECT customerid,username,'x' as password,uid,gid,'Froxlor User' as comment,homedir,shell, login_enabled FROM ftp_users ORDER BY uid ASC;
results in userftp1 userftp2 userftp3 being listed before the user
SELECT customerid,username,'x' as password,uid,gid,'Froxlor User' as comment,homedir,shell, login_enabled FROM ftp_users ORDER BY id ASC;
results in user listed before userftp1, userftp2, userftp3.
This fixes the problem with the ownership.

# Description

Please include a summary of the change and which issue is fixed if any. Please also include relevant motivation and context. List any dependencies that are required for this change.

Fixes # (issue)

## Type of change

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

- [x] Test A
- [x] Test B

**Test Configuration**:

* Distribution:
* Webserver:
* PHP:
* etc.etc.:

# Checklist:

- [x] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/lib/Froxlor/Cron/System/Extrausers.php
+++ b/lib/Froxlor/Cron/System/Extrausers.php
@@ -24,7 +24,7 @@ class Extrausers
 	{
 		// passwd
 		$passwd = '/var/lib/extrausers/passwd';
-		$sql = "SELECT customerid,username,'x' as password,uid,gid,'Froxlor User' as comment,homedir,shell, login_enabled FROM ftp_users ORDER BY uid ASC";
+		$sql = "SELECT customerid,username,'x' as password,uid,gid,'Froxlor User' as comment,homedir,shell, login_enabled FROM ftp_users ORDER BY id ASC";
 		self::generateFile($passwd, $sql, $cronlog);
 
 		// group


d00p:
Just cosmetics, but we could keep the sort by `uid` and within that sort by `id` (so it keeps the same customer together in the passwd list). As said, pure cosmetics. And thanks for noticing :)

tokijici:
Sorting by UID and ID is also fine, but not cosmetic change only as it affects the ownership of the files, the are within the same UID but still not right. Many thanks!

d00p:
How can the order of the SAME user (uid) affect ownership? Well, on all boxes I've tried to reproduce this, I am unable to do so as I'm always getting the 'main' user before the others.

How about something like that:

```sql
ORDER BY uid, LENGTH(username) ASC
```

tokijici:
I get the userftp1 to be the owner of the user directory instead of just user.
The system got migrated to a new server - ubuntu 20.04, many users have been removed along the way so the database state might hard to replicate in a testing environment, the proposed query works just es expected and returns the right user order on my server, so the entire query will be:
`SELECT customerid,username,'x' as password,uid,gid,'Froxlor User' as comment,homedir,shell, login_enabled FROM ftp_users ORDER BY uid, LENGTH(username) ASC;`
Cheers!

PR #949, opened 2021-06-23
Custom CSS File in default theme

bashgeek:
Me again ;-)

Another small thing, this enables you to put a "custom.css" file into the template folder (templates/Sparkle/assets/css/custom.css), analog to the logo_custom.png capability. This will be added after the actual stylesheets of the theme and allows to override certain things like colors pretty easily without the need to have to maintain your own full theme just for some color or font changes.

Again, this is not a breaking-change, if the file doesn't exist, it's simply ignored (this logic already existed before).


d00p:
Also good idea, I'd suggest to add the custom.css file already with a header-comment e.g. `/* put your custom css here */` or similar. Makes it easier to see that this possibility exists

bashgeek:
@d00p That would be good, but that will result in having your custom.css file overwritten/flushed to that default with every Froxlor update, wouldn't it? :D

I added a "custom.example.css" file in that folder now with instructions to rename it to "custom.css" and put CSS in there. Sounds good?

d00p:
You're right yes. Think this is a good solution then.

PR #956, opened 2021-07-18
Feature-request #672 - database name prefixes + custom name

### LINKED ISSUE #672: [feature request] database name prefixes + custom name
I would like to choose the names for my databases.

I have now 7 databases called mostly like

user1
user2
...
user7

Is difficult to know which one is doing what
thx.

monster010:
# Description
After I saw that the pull request #680 was no longer changed, I decided to do the whole thing myself.
But I implemented the whole thing a little differently.

Add mysql prefix option to user customername + database name field as databasename.

Fixes #672 

## Type of change

- [x] This change requires a documentation update

# How Has This Been Tested?
Locally and in an active environment. If the newely option value "DBNAME" is used, new functionality is triggered and executed.

# Checklist:


lib/Froxlor/Database/DbManager.php
```diff
@@ -87,6 +87,8 @@ public function createDatabase($loginname = null, $password = null, $last_accnum
 			while (in_array($username, $allsqlusers)) {
 				$username = $loginname . '-' . substr(md5(uniqid(microtime(), 1)), 20, 3);
 			}
+		} elseif (strtoupper(Settings::Get('customer.mysqlprefix')) == 'DBNAME') {
```
d00p, line 90: This would lead to the same username being used for multiple databases which might be an unwanted behaviour. An idea would be to use the random-username-generation-code from the above case (mysqlprefix == 'RANDOM') so each database has its unique username

d00p (changes requested):
As this PR changes/adds functions to API (`Mysqls.add()`) it would be very nice to have this case covered in the corresponding PHPUnit test, see `tests/Mysqls/MysqlsTest.php`

lib/Froxlor/Database/DbManager.php
monster010: This is not correct because the user that is being created corresponds to the name of the database. Each database has its own user who has the same name as the database itself.

lib/Froxlor/Api/Commands/Mysqls.php
```diff
@@ -79,7 +81,12 @@ public function add()
 		);
 		// create database, user, set permissions, etc.pp.
 		$dbm = new \Froxlor\Database\DbManager($this->logger());
-		$username = $dbm->createDatabase($newdb_params['loginname'], $password, $newdb_params['mysql_lastaccountnumber']);
+
+		if(strtoupper(Settings::Get('customer.mysqlprefix')) == 'DBNAME' && !empty($databasename)) {
+			$username = $dbm->createDatabase($newdb_params['loginname'].'_'.$databasename, $password);
```
monster010, line 86: At this point I have to replace the hard-coded underline.

lib/Froxlor/Api/Commands/Mysqls.php
```diff
@@ -79,7 +83,12 @@ public function add()
...
+		} elseif(empty($databasename)) {
```
d00p: just use `} else {` because `$databasename` will never be empty

tests/Mysqls/MysqlsTest.php
```diff
@@ -52,6 +52,36 @@ public function testCustomerMysqlsAdd()
...
+		$newPwd = \Froxlor\System\Crypt::generatePassword();
+		$data = [
+			'mysql_password' => $newPwd,
+			'database_name' => 'abc123',
+			'description' => 'testdb',
+			'sendinfomail' => TRAVIS_CI == 1 ? 0 : 1
+		];
+		$json_result = Mysqls::getLocal($customer_userdata, $data)->add();
+		$result = json_decode($json_result, true)['data'];
+		$this->assertEquals('abc123', $result['databasename']);
```
d00p: this will fail as the generated database-name equals `test1_abc123`

lib/Froxlor/Api/Commands/Mysqls.php
d00p: why? it's an allowed character and makes total sense if the name is prefixed with the customer's loginname

lib/Froxlor/Api/Commands/Mysqls.php
```diff
@@ -31,13 +31,15 @@ class Mysqls extends \Froxlor\Api\ApiCommand implements \Froxlor\Api\ResourceEnt
 	 *        	optional, default is 0
 	 * @param string $description
 	 *        	optional, description for database
+	 * @param string $name
```
d00p, line 34: the variable-name `$name` should be equal the `getParam('variable-name')` as the automated API documentation takes its info of the phpDoc.

Also `database_name` could be irritating, maybe we should use something like `custom_suffix`? 

phpDoc should then be
`@param string $custom_suffix`
and to get the parameter:
`$databasename = $this->getParam('custom_suffix', true, '');`

d00p (changes requested):

tests/Mysqls/MysqlsTest.php
d00p: resolved in current commit

monster010 (commented):
I am currently thinking about how to incorporate the tests in the most sensible way.

lib/Froxlor/Api/Commands/Mysqls.php
monster010: I changed it to custom_suffix =)

d00p:
> I am currently thinking about how to incorporate the tests in the most sensible way.

Well currently the one that is failing is just the listing test that expects one entry for the customer but due to your test now has two...either remove your test-database after successful test or increase the expected number of databases in the listing test ;)

d00p:
Thanks a lot, nice work

PR #959, opened 2021-07-20
Fix Minimum and Expired SOA-Records

ndum:
Currently in $soa_content the "Minimun SOA record" is obtained from the system configuration -> TTL and thus may be incorrect. (TTL and Minimum SOA-Records are not the same?! looks like a bug)

In addition, I also fixed the "Expired" entry, according to current [RFC](https://www.ietf.org/rfc/rfc1912.txt) this must be 2 or 4 weeks.

Thanks for merging.


d00p:
Sounds good ;)

PR #974, opened 2021-08-18
Adding support for PowerDNS-Replication

ndum:
Hi

Currently, "NATIVE" is always entered into the database as the default PowerDNS type-field. However, if a PowerDNS Master/Slave replication is required, "MASTER" is needed as type.

I have now added another option in the settings. There you can choose between "Native" (Default - as before) and "Master". 

DB update-files and language-files are created and tested.

Thanks for merging!

Unbenannt123


install/froxlor.sql
```diff
@@ -611,6 +611,7 @@ opcache.interned_strings_buffer'),
 	('system', 'documentroot_use_default_value', '0'),
 	('system', 'passwordcryptfunc', '3'),
 	('system', 'axfrservers', ''),
+	('system', 'powerdns_mode', 'Native'),
```
d00p, line 614: You also need to adjust the db_version to `202108180` in this file for new installations

lib/Froxlor/Cron/Dns/PowerDNS.php
```diff
@@ -111,9 +113,15 @@ private function clearZoneTables($domains = null)
 
 	private function insertZone($domainname, $serial = 0)
 	{
-		$ins_stmt = \Froxlor\Dns\PowerDNS::getDB()->prepare("
+		if (Settings::Get('system.powerdns_mode') === 'Master') {
```
d00p, line 116: Why not just making `type` also a variable for the prepared statement?

```php
$ins_stmt = \Froxlor\Dns\PowerDNS::getDB()->prepare("
    INSERT INTO domains set `name` = :domainname, `notified_serial` = :serial, `type` = :type
");
$ins_stmt->execute(array(
	'domainname' => $domainname,
	'serial' => $serial,
        'type' => strtoupper(Settings::Get('system.powerdns_mode'))
));
```

d00p (changes requested):

install/froxlor.sql
ndum: OK - thanks for the hint. It is now implemented in: https://github.com/Froxlor/Froxlor/pull/974/commits/29586bd29c2ff4e9b01ac34e57143a353e9aceb3

lib/Froxlor/Cron/Dns/PowerDNS.php
ndum: Thanks for the hint here too! Its fixed also in: https://github.com/Froxlor/Froxlor/pull/974/commits/29586bd29c2ff4e9b01ac34e57143a353e9aceb3

d00p (approved):

PR #976, opened 2021-09-03
[WIP] czech.lng.php

rex2630:
I tested it after it got released to public and it doesnt work, is there any file to change for froxlor to support it? And in this PR I'm continuing to work on this lang.


rex2630:
How to prepare strings for release 0.11? Lot of strings seems to be missing in the dev branch.

d00p:
> How to prepare strings for release 0.11? Lot of strings seems to be missing in the dev branch.

same way as always....just the filenames are a bit different. English (en.lng.php) is the default/base. Everthing in there needs to be/should be translated in the target language-file.

d00p:
Is still still being worked on?

d00p:
no feedback

PR #979, opened 2021-10-07
feat: add mysql tls support

NickUfer:
# Description

We at INWX required Froxlor to work with our new database proxy which only handles encrypted traffic. For that we created a patch which we now want to contribute to the main project.
This PR adds the option to use TLS for connecting to MySQL with or without server certificate validation. 

TLS is enabled when a server certificate is provided. The option to verify the server certificate is by default disabled and only has an effect when a server certificate is provided.

## Type of change

# How Has This Been Tested?

I tested the installation process locally from begin to finish multiple times with a simple functional test afterwards. A colleague in system administration tested the change too.

- Test 1: Without provided certificate and no certificate validation
- Test 2: With provided certificate and no certificate validation
- Test 3: With provided certificate and certificate validation

**Test Configuration**:

* Distribution: Kubuntu 21.04
* Webserver: Nginx
* PHP: PHP7.4
* MariaDB: 10.6 configured with TLS with certificates self signed with `cfssl`.

# Checklist:

- [X] New and existing unit tests pass locally with my changes


d00p:
Very very nice, thank you for the contribution.

schadom:
Thanks so much @NickUfer! We had been looking for the same. For the info - if `mysql_ssl_ca_file` is set to "true" and `mysql_ssl_verify_server_certificate` is set to "false" TLS is still active but without certificate authentication. See https://stackoverflow.com/a/56153645

PR #1008, opened 2022-01-26
Improve composer.json syntax

MrKrisKrisu:
# Description

This change improves the syntax in composer.json.

1. `php` was listed twice (in `require` and `require-dev`).
2. `>=` syntax should be avoided to account for breaking changes in major PHP versions. `^` allows minor updates, but not major

(3. I would like to see the minimum version raised to PHP 7.4, as 7.1 and 7.2 are no longer updated and should not be supported in my opinion. **Can I also implement this in this PR?**)

## Type of change

# How Has This Been Tested?

--

**Test Configuration**:

--

# Checklist:

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -30,7 +30,7 @@
 		"docs": "https://github.com/Froxlor/Froxlor/wiki"
 	},
 	"require": {
-		"php": ">=7.1",
+		"php": "^7.1 || ^8.0",
 		"ext-session": "*",
 		"ext-ctype": "*",
 		"ext-pdo": "*",
@@ -52,7 +52,6 @@
     },
 	"require-dev": {
 		"phpunit/phpunit": "^9",
-		"php": ">=7.3",
 		"ext-pcntl": "*",
 		"phpcompatibility/php-compatibility": "*",
 		"squizlabs/php_codesniffer": "*",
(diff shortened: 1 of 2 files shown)


d00p:
> (3. I would like to see the minimum version raised to PHP 7.4, as 7.1 and 7.2 are no longer updated and should not be supported in my opinion. Can I also implement this in this PR?)

This is discussed very often and I'm totally with you on this but from my experience, there are too many users who are still on older distributions and only stick with the default php (which e.g. for debian buster is 7.3).
As long as the source does not really require a higher php version, there is no reason to raise the requirement (development point of view).

MrKrisKrisu:
Okay! That's why i've asked before just doing it in the PR. :)

Do you agree with the rest of the change?

d00p:
basically yes, but from a brief overview of the lock file, can we be sure that we do not break compatibility with php-7.1? (not dev-environment of course)

MrKrisKrisu:
The only php version changes in the lock file are from:

- `phpspec/prophecy` up to `^7.2` 
    - only used in require-dev

- `psr/container` from `>=7.2` to `>=7.4` 
    - used by `symfony/dependency-injection`
    - used by `pdepend/pdepend`
    - used by `phpmd/phpmd`
    - only used in require-dev

So I did not find any version changes in the lockfile. Only dev-reqs has been changed to a minimum of PHP 7.4.

PR #1021, opened 2022-04-10
Create update-database.php

### LINKED ISSUE #1015: Unattended installation
Is there a way to complete installation from cli

I wan't to automatically install/configure my servers with saltstack.
Is there a way to create all config files, the upgrade process and installation from cli

php /var/www/froxlor/scripts/installation.php --configFile=/etc/froxlor/config.json

and in the config json are the paramaters needed for installation like mysql root password and so on?

Kapsonfire-DE:
allows to execute the upgrade process via clia

Fixes #1015 

# Checklist:

- [X ] I have commented my code, particularly in hard-to-understand areas

### DIFF
--- a/install/scripts/update-database.php
+++ b/install/scripts/update-database.php
@@ -1,0 +1,23 @@
+<?php
+// Include autoloader and table constants
+require __DIR__.'/../../vendor/autoload.php';
+require \Froxlor\Froxlor::getInstallDir() . '/lib/tables.inc.php';
+
+// remove html for console output
+function cleanUpdateOutput($buffer) {
+        static $newLine = true;
+        $buffer = strip_tags(preg_replace("/<br\W*?\/>/", "\n", $buffer));
+        if($newLine) {
+                $buffer = date('[H:i:s] ').$buffer;
+        }
+
+        $newLine = strpos($buffer, "\n") !== false              ;
+        return $buffer;
+}
+
+define('_CRON_UPDATE', 1);
+ob_start('cleanUpdateOutput', 2); //force direct flushing
+
+// execute update scripts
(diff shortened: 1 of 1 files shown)


d00p:
weren't you working on an "unattended installation" script? Thought this was just part of it...

install/scripts/update-database.php
```diff
@@ -0,0 +1,23 @@
+<?php
```
d00p, line 1: could prefix `#!/usr/bin/php`

install/scripts/update-database.php
```diff
@@ -0,0 +1,23 @@
...
+// Include autoloader and table constants
```
d00p, line 2: please add this to avoid calling this via web directly (as it does not feature any authentication whatsoever)

```php
// Check if we're in the CLI
if (@php_sapi_name() !== 'cli') {
	die('This script will only work in the shell.');
}
```

d00p (changes requested):

Kapsonfire-DE:
The update-database is needed for unattended updates, but not the installation.

I've done the unattented installation via saltstack using pillars for credentials and so on. Not sure if this should be part of the Froxlor Repo. Probably I'm creating a saltstack formula for it.

d00p:
I thought there would be something like `/install/install-cli.php` with an additional parameter `--update` for updating via CLI...to have a complete CLI solution

Kapsonfire-DE:
I can do this - should be in style like config-services to create a json config first and after that applying that config file?

d00p:
Maybe the best would be to join the development channel on our discord server. There is a lot going on for the next major release - I think it would make much more sense to do a thing like that for the next version

PR #1028, opened 2022-04-18
Add global acme reload command (#822)

### LINKED ISSUE #822: [feature request] letsencrypt: add post issue/renewal hook possibility
Requesting a feature to add a possibility to run scripts/code/commands after a letsencrypt issue/renewal was running.

At least pointing to a shell script (file) should be enough.

Background:
As acme.sh runs as root by the froxlor cron, all certificates and keys are owned by root and filemode is 0600. If you maybe manually changed owner or filemode of the certs - each run of acme.sh resets owner and filemode. This is by design of acme.sh and for security reasons the right way.

Adding the requested feature could make it possible to allow further usage of the certificates. 
Advanced system engineers could copy the/certain files and make necessary recurring changes needed after every renewal/issue.

Usecase:

On my Hosts I run different daemons. Some are not running as root but also need ssl-certificates. non-root daemons/users cannot access the certificate files stored by scme.sh.
Thus I currently manually copy and chown the needed certs after they renewed. This would be nice to happen automatically. 

Sadly, also acme.sh does not provide a similar posthook - and even if - currently you could not use it in froxlor.

FliegenKLATSCH:
# Description

Add a setting to pass the --reloadcmd to acme.sh

Fixes #822

## Type of change

### DIFF
--- a/actions/admin/settings/131.ssl.php
+++ b/actions/admin/settings/131.ssl.php
@@ -142,6 +142,14 @@ return array(
 					'default' => '/root/.acme.sh/acme.sh',
 					'save_method' => 'storeSettingField'
 				),
+				'system_acmereloadcmd' => array(
+					'label' => $lng['serversettings']['acmereloadcmd'],
+					'settinggroup' => 'system',
+					'varname' => 'acmereloadcmd',
+					'type' => 'string',
+					'default' => '',
+					'save_method' => 'storeSettingField'
+				),
 				'system_letsencryptacmeconf' => array(
 					'label' => $lng['serversettings']['letsencryptacmeconf'],
 					'settinggroup' => 'system',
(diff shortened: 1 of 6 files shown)


d00p:
Shouldn't that be domain-specific not globally? In your example, dovecot and postfix would get restarted for EVERY certificate renew of every domain

snoopotic:
Yeah, I also reuse one of my LE certificates for the mailserver. 
Maybe there can be a setting: "used for Mailserver" and only if this cert gets renewed, mailserver services get restarted.

d00p:
pretty specific that way...what if i have some other services running with one of my froxlor-administrated domains and wanted to reload these after renew? 
But dangerous, as i would NEVER allow a customer to enter a reload command to be run on the server after certificate renew.

FliegenKLATSCH:
I agree, this setting should not be visible to any customer. Are those serversettings visible to them?
The proposed description says that it is executed for each certificate. For me that's fine, the reload doesn't hurt and I don't have that many domains..
If someone wants to restrict this, it could be possible with ```[[ "${Le_Domain}" == "example.com" ]] && systemctl reload postfix.service``` in that setting (see https://github.com/acmesh-official/acme.sh/blob/master/acme.sh#L5867)

Issue #515, opened 2018-02-13
Add Option to set OpenBaseDir manually for domain/subdomain


dirkey:
The OpenBaseDir can only be set to the "Home Directory" or the "Document Root". It would be nice to have a third option to set it manually. For example: 
Magento needs only it's pub directory to be accessible via Apache (DocumentRoot), but PHP needs access to all files in the "Mangento Root" (OpenBaseDir). Currently I must give it access to all files under the home directory of the user.

d00p:
So, domain-documentroot is e.g. **/[user]/magento/pub** and your chice would be either **/[user]/** or **/[user]/magento/pub** but you need **/[user]/magento** - correct?

d00p:
One of the "minor" problems here is that we try to avoid letting the customer enter paths manually as it might void the openbasedir functionality which the admin might wanted intentionally..that's why it's only these two choices

dirkey:
I see! Maybe, we let users only allow to add relative paths to there home directory. But then, we need to take care of relative paths pointing to a parent directory. A solution is also to show the directory tree instead of a textbox. Then he is only allowed to choose these.

NightmareJoker2:
My two cents: I'd love this. I've been creating separate users and using their home directories for it and that's become quite tedious to manage.
Allow the following three options: 
* document root
* parent of document root (must be inside home directory)
* home directory

If the parent of the document root is not inside the home directory, automatically set the home directory. (It's an absolute path string, so something like 
```
$home_path = realpath($home_directory);
$open_basedir_path = realpath($open_basedir);
if ($home_path === $open_basedir_path || substr($open_basedir_path, 0, strlen($home_path) + 1) === $home_path.DIRECTORY_SEPARATOR)
{
    // everything is fine
}
else
{
    $open_basedir_path = $home_path;
}
```
should suffice to make this safe.

This would help greatly with properly setting up Drupal, Magento, Tyop3 and custom built applications, such that only one small `index.php` and static files remain in the document root of the webs server's virtual host, and all application files, view templates (typically not with `php` extension, would be served as text), configuration, and possibly databases (e.g. SQLite ones) can safely remain outside of it and out of the reach of a FastCGI or rewrite engine failure of some sort from a package upgrade mishap.

d00p:
Sounds good to me @NightmareJoker2

florianschroen:
[...]

PR #1031, opened 2022-05-08
[WIP] Disable default 'root' directive in nginx

patschi:
# Description

These two rather simple changes solved two very specific use-cases for me. I had these two workarounds in place hard-coded, but I thought it might be beneficial and allowing more flexibility when implementing them in the UI for everyone.

image

I'm running a a forum system which needs the folder `public/` to be set as the webroot, as it contains the public-facing PHP files there. However when setting webroot to `/var/customers/webs/user/domain.tld/public/` the website fails as `open_basedir` uses the webroot path, blocking access to required PHP files one folder up.

Currently I have set webroot as `/var/customers/webs/user/domain.tld/` in Froxlor and the default `root` directive disabled, and using `root /var/customers/webs/user/domain.tld/public/` in the custom vhost configuration.

## Type of change

# How Has This Been Tested?

- [X] Disabled both options on above screenshot and run cronjob, the automated `root` and `location / {}` blocks are not added.
- [X] By default the values are enabled, not affecting existing installations.
- [ ] Tested the update or installation of Froxlor with the new database changes (not selected on purpose)

**Test Configuration**:

* Distribution: Ubuntu 20.04
* Webserver: nginx
* PHP: 8.0

# Checklist:


d00p:
Sounds rather like a way to use a "hack" to get around open_basedir in a weird way.
I think the root-issue here is what is discussed here: https://github.com/Froxlor/Froxlor/issues/515

d00p:
froxlor-2.0 has the option to set  the openbasedir to the parent-directory of the domain-documentroot, exactly what you've described

PR #1125, opened 2023-04-19
Allow admins to edit openbasedir_path for domains

### LINKED ISSUE #1086: OpenBasedir-path can't be set via API for apex domains
**Describe the bug**
The API offers the parameter `openbasedir_path`  for `SubDomains.add`, but the same parameter is missing for `Domains.add`. Via the Admininterface, also `openbasedir_path` can't be set or edited, only the customer can set it via the Webinterface.

**System information**
* Froxlor version: 2.0.8
* Web server: apache2
* DNS server: none
* POP/IMAP server: Dovecot
* SMTP server: postfix
* FTP server: proftpd
* OS/Version: Debian 11

**To Reproduce**
The missing of the mentioned parameter for `Domains.add` is documented here in the official [API-Docs](https://docs.froxlor.org/latest/api-guide/commands/domains.html#domains-add).

The mentioned parameter for `SubDomains.add` is documented here in the official [API-Docs](https://docs.froxlor.org/latest/api-guide/commands/subdomains.html#subdomains-add).

The following screenshot shows the option for an apex domain:
openbasdirpath

**Expected behavior**
The parameter `openbasedir_path` should be available for `Domains.add` and `Domains.update`.

**Additional context**

* Im not actually sure if this is a bug. If it isn't, this shall be seen as feature request.

lukasbableck:
# Description
Admins should be able to modify the OpenBasedir path for domains, just like customers can.

Fixes #1086

## Type of change

# How Has This Been Tested?

I have not added unit tests for this, as there are no tests for the openbasedir field or the openbasedir_path field for Subdomains either, but I have tested it manually by adding and updating domains while changing this field.

**Test Configuration**:

* Distribution: Debian 11
* Webserver: nginx 
* PHP: 8.2.5 FPM
* etc.etc.:

# Checklist:

### DIFF
--- a/lib/Froxlor/Api/Commands/Domains.php
+++ b/lib/Froxlor/Api/Commands/Domains.php
@@ -225,6 +225,8 @@ class Domains extends ApiCommand implements ResourceEntity
 	 *            optional, whether php is enabled for this domain, default 0 (false)
 	 * @param bool $openbasedir
 	 *            optional, whether to activate openbasedir restriction for this domain, default 0 (false)
+	 * @param int $openbasedir_path
+	 *            optional, either 0 for domains-docroot, 1 for customers-homedir or 2 for parent-directory of domains-docroot
 	 * @param int $phpsettingid
 	 *            optional, specify php-configuration that is being used by id, default 1 (system-default)
 	 * @param int $mod_fcgid_starter
@@ -312,6 +314,7 @@ class Domains extends ApiCommand implements ResourceEntity
 				$documentroot = $this->getParam('documentroot', true, '');
 				$phpenabled = $this->getBoolParam('phpenabled', true, 0);
 				$openbasedir = $this->getBoolParam('openbasedir', true, 0);
+				$openbasedir_path = $this->getBoolParam('openbasedir_path', true, 0);
 				$phpsettingid = $this->getParam('phpsettingid', true, 1);
 				$mod_fcgid_starter = $this->getParam('mod_fcgid_starter', true, -1);
 				$mod_fcgid_maxrequests = $this->getParam('mod_fcgid_maxrequests', true, -1);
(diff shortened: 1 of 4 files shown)


lib/Froxlor/Api/Commands/Domains.php
```diff
@@ -312,6 +314,7 @@ public function add()
 				$documentroot = $this->getParam('documentroot', true, '');
 				$phpenabled = $this->getBoolParam('phpenabled', true, 0);
 				$openbasedir = $this->getBoolParam('openbasedir', true, 0);
+				$openbasedir_path = $this->getBoolParam('openbasedir_path', true, 0);
```
d00p, line 317: `getBoolParam()` will only give your values `1` or `0` never the value `2` which is also possible for this setting

lib/formfields/admin/domains/formfield.domains_edit.php
```diff
@@ -390,6 +390,13 @@
 						'value' => '1',
 						'checked' => $result['openbasedir']
 					],
+					'openbasedir_path' => [
+						'visible' => $result['openbasedir'] == '1',
```
d00p, line 394: what if the value for `openbasedir` is to be changed when editing and then the admin can only see this option when entity is saved and reopened

d00p (changes requested):

lib/formfields/admin/domains/formfield.domains_edit.php
lukasbableck: True, it's better when the field is always visible

lib/Froxlor/Api/Commands/Domains.php
lukasbableck: Fixed

PR #1180, opened 2023-09-05
set autofocus on search if no menu item is opened

### LINKED ISSUE #1178: [FR] auto focus on the search on dashboard
right after login it would help, if the cursor is automatically focussed in the Search Input Field.

alternatively a keyboard-shortcut on the search field would help also, but only, if this doesn't interfere with other input-fields on some pages

rubo77:
The cursor is automatically focused in the Search Input Field if you are on the dashboard, where no menu item is opened.

closes #1178


envoyr:
I'm against this feature, autofocusing the cursor over the search every time isn't always helpful. This makes sense for a login or register, but not necessarily for the search.

rubo77:
@envoyr  I am not sure if you were aware, that I only wanted to autofocus the search on the dashboard once.

But ok

PR #1181, opened 2023-09-05
ignore package-lock.json

rubo77:
This file should not be within the git repository, since it is automatically edited, when you run the install instructions:

```
composer install --no-dev
npm install
npm run build
```


d00p:
same as composer.lock, it represents the package versions installed/tested for a release and thus used for building the release packages (instead of package.json).

Also relevant for developers: It allows future devs to install the same dependencies in the project.

rubo77:
Ah ok. So would you like to see the changes to the file, that were automatically made in my repository, and update the file then?

d00p:
You obviously don't understand and rather discuss meaningless on GitHub then talk to us on discord.

discordier:
For the record, install as follows:
```
composer install --no-dev
npm ci
npm run build
```
See https://docs.npmjs.com/cli/v9/commands/npm-ci

That way you get the dependencies as intended.

`npm install` is for developers that want to get current dependencies.

PR #1189, opened 2023-09-17
Add focus to global search clicked right outside of the input field

rubo77:
This also focuses, if the viewport is so narrow, that it is switched to the mobile-view


rubo77:
please reopen, I force pushed a rebase

PR #1208, opened 2023-12-16
Add manual_config install var to cli

Gamerboy59:
Make the manual_config var, which is available to the web installer, usuable for the cli installer too. If manual_config is set to true skip else (not set or false) proceed with auto config.

Maybe example in docs needs to be updated, too.

### DIFF
--- a/lib/Froxlor/Cli/InstallCommand.php
+++ b/lib/Froxlor/Cli/InstallCommand.php
@@ -273,7 +273,7 @@ final class InstallCommand extends Command
 					$cmdfield['label'],
 					$cmdfield['value']
 				]);
-				if (!empty($decoded_input) || $this->io->confirm('Execute command now?', false)) {
+				if (!empty($decoded_input) ? (!isset($decoded_input['manual_config']) || !filter_var($decoded_input['manual_config'], FILTER_VALIDATE_BOOLEAN)) : $this->io->confirm('Execute command now?', false)) {
 					passthru($cmdfield['value']);
 				}
 				break;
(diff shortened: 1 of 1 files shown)


d00p:
Well the CliInstaller ist noch 100% exactly like the WebInstaller ...if you skip the configuration step, the webinstaller will generate the required lib/userdata.inc.php file (instead of doing that with the command `bin/froxlor-cli froxlor:install -c`

So if you want to skip the services configuration within the cli install, you'd still need to create the `lib/userdata.inc.php` file

Gamerboy59:
> Correction: the cliinstaller already creates the userdata.inc.php on step 3 - so i guess it's fine to skip the configuration

I'm not quite sure if I understand your correction statement correctly. Is it fine to skip the config like in this PR or do you want it differently or is it already skipped?

d00p:
I was kind of thinking out loud sorry if it was a bit confusing.

No, basically it's all fine, I would do it a bit differently, don't really like a ternary operator within an if-statement. Something like this:

```diff
-                               if (!empty($decoded_input) || $this->io->confirm('Execute command now?', false)) {
-                                       passthru($cmdfield['value']);
+                               if (!empty($decoded_input) &&
+                                       (!isset($decoded_input['manual_config']) || (bool)$decoded_input['manual_config'] === false)
+                               ) {
+                                       if ($this->io->confirm('Execute command now?', false)) {
+                                               passthru($cmdfield['value']);
+                                       }
                                }
```

what do you think?

Gamerboy59:
We can do it without ternary but in your example the confirm dialog would always appear which is not good for fully unattended runs. The passthru cmd should be executed even if the `manual_config` variable is missing or false and don't cause a prompt. The prompt should only appear if there `$decoded_input` is emtpy.

d00p:
Yes, you're right, this should do:

```php
				if (!isset($decoded_input['manual_config']) || (bool)$decoded_input['manual_config'] === false) {
					if (!empty($decoded_input) || $this->io->confirm('Execute command now?', false)) {
						passthru($cmdfield['value']);
					}
				}
```

1) check for "manual_config" to be unset or at least "false", otherwise dont run anything
2) If an input file was given OR the confirmation was answered, run command

Gamerboy59:
Yes thanks, this looks like a disassembling of the ternary version.  
I've updated the PR.

PR #1209, opened 2023-12-17
Fix auto config cmd display on manual_config

Gamerboy59:
Only display command for automatic install if no manual_config was specified.

_TODO: It would be good to only display the auto cmd if the "Execute command now?" confirmation request pops up._

### DIFF
--- a/lib/Froxlor/Cli/InstallCommand.php
+++ b/lib/Froxlor/Cli/InstallCommand.php
@@ -265,15 +265,15 @@ final class InstallCommand extends Command
 				return $this->showStep(++$step, $extended, $decoded_input);
 				break;
 			case 4:
-				$section = $inst->formfield['install']['sections']['step' . $step] ?? [];
-				$this->io->section($section['title']);
-				$this->io->note($this->cliTextFormat($section['description']));
-				$cmdfield = $section['fields']['system'];
-				$this->io->success([
-					$cmdfield['label'],
-					$cmdfield['value']
-				]);
 				if (!isset($decoded_input['manual_config']) || (bool)$decoded_input['manual_config'] === false) {
+					$section = $inst->formfield['install']['sections']['step' . $step] ?? [];
+					$this->io->section($section['title']);
+					$this->io->note($this->cliTextFormat($section['description']));
+					$cmdfield = $section['fields']['system'];
+					$this->io->success([
+						$cmdfield['label'],
+						$cmdfield['value']
+					]);
 					if (!empty($decoded_input) || $this->io->confirm('Execute command now?', false)) {
 						passthru($cmdfield['value']);
 					}


d00p:
I don't really get your english sentences, please try to describe more detailed what your goal is.

Why would you not display the commands if manual_config is true...maybe someone wants to copy'n'paste it for later use or just adjust them a bit...also, it doesnt hurt at all, nothing gets created.

> TODO: It would be good to only display all if the "Execute command now?" confirmation request pops up.

What? No. Please explain, no idea what you mean or what the goal of this should be

Gamerboy59:
The idea is that if the `manual_config` var is predefined to configure the system manually, there is no point in displaying the command for automatic configuration because that is not what the user wanted. Why would he specify the `manual_config` variable if he is looking for automatic configuration?

It doesn't "hurt" but can be confusing if ending statement tells you to run a certain command. Well yes, if a user wants to adjust the command, it can be useful. However, the user can obtain the command again at any time in the web panel if he changes his mind and wants a `manual_config: false` setup.

d00p:
The `manual_config` flag comes just from the webui installation so you can skip the "waiting for services to be configured" and get redirected to the login. It has no real purpose for the cli installation. I think you misunderstood its purpose. It's all about the finishing commands. It is not meant to be used. I agree it made sense in your prior PR but that's about the influence its got. You can now decide whether or not the finishing commands are being executed **if using a file input** (e.g. unattended install). For every other case, the installation is just fine and does exactly what it's supposed to.

Gamerboy59:
> I think you misunderstood its purpose. It's all about the finishing commands. It is not meant to be used.

Ok. I was looking to make it behave like an unattended installation where you usually are not offered anything to execute in a later step. Just a result like unattended installation was successful or not.

d00p:
> behave like an unattended installation where you usually

Did you actually test it? That's exactly what the cli install command does if you provide an input file
...

Gamerboy59:
Yes, it finishes with success or not but at the end there's this wall of text which highlights actions I could execute as following step.
I was expected something like "installation finished as given in input-file" and the process status code and not more. It's just my understanding of an unattended installation design though, that's why I brought up the idea but we can leave it as it is without functional disadvantages.

d00p:
Sorry, no idea what you mean. If I run the installation command with and Input file, I don't have to do anything and have a fully installed and configured froxlor system - which is my understanding of a complete unattended setup.

If I don't use any input file, I get asked questions about the Infos required and in the end, it will be executed and I also have a completely setup system. 

So where exactly is the issue now?

Gamerboy59:
As said, this isn't about functional/technical issues, it's more about the design. If I run something in unattended mode where I explicitely state that I want to configure something manually, I don't need a hint how to do it automatically regardless. It doesn't bother though.

d00p:
>  I don't need a hint how to do it automatically regardless.

you said it yourself, YOU dont need a hint, others might

PR #1228, opened 2024-01-16
Check for Argon2 support before using constant PASSWORD_ARGON2X

sro0:
# Description

Adding a new mail account on a system without PHP Argon2X support an exception will be triggered.

e1

-> Check if Argon2X is supported in PHP (==constant is defined) before using it.

## Type of change

# How Has This Been Tested?

Adding a mail account on a system without PHP Argon2X support was possible after patch

### DIFF
--- a/lib/Froxlor/Api/Commands/EmailAccounts.php
+++ b/lib/Froxlor/Api/Commands/EmailAccounts.php
@@ -157,10 +157,10 @@ class EmailAccounts extends ApiCommand implements ResourceEntity
 
 			// prefix hash-algo
 			switch (Settings::Get('system.passwordcryptfunc')) {
-				case PASSWORD_ARGON2I:
+				case defined('PASSWORD_ARGON2I') && PASSWORD_ARGON2I:
 					$cpPrefix = '{ARGON2I}';
 					break;
-				case PASSWORD_ARGON2ID:
+				case defined('PASSWORD_ARGON2ID') && PASSWORD_ARGON2ID:
 					$cpPrefix = '{ARGON2ID}';
 					break;
 				default:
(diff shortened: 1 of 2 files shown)


d00p:
The API in _EmailAccounts.add()_ uses the setting `system.passwordcryptfunc` to determine the hash to be used. This setting, when opening the security settings, gets all available password-hashes (see https://github.com/Froxlor/Froxlor/blob/v2.1/actions/admin/settings/210.security.php#L53) , which uses Crypt::getAvailablePasswordHashes() (see https://github.com/Froxlor/Froxlor/blob/v2.1/lib/Froxlor/System/Crypt.php#L97) which only adds available hashes to the select for the setting

So, most likely a hash-algorithm is/was set which became unavailable.

d00p:
Here's a more "systemwide"-safe solution:

```diff
diff --git a/lib/Froxlor/Settings.php b/lib/Froxlor/Settings.php
index dad81814..056ed8ed 100644
--- a/lib/Froxlor/Settings.php
+++ b/lib/Froxlor/Settings.php
@@ -26,6 +26,7 @@
 namespace Froxlor;

 use Froxlor\Database\Database;
+use Froxlor\System\Crypt;
 use PDO;
 use PDOStatement;

@@ -159,7 +160,17 @@ class Settings
                $result = null;
                if (isset(self::$data[$sstr[0]][$sstr[1]])) {
                        $result = self::$data[$sstr[0]][$sstr[1]];
+
+                       // if password-crypt-func is requested, check whether it's (still) available
+                       if ($sstr[0] == 'system' && $sstr[1] == 'passwordcryptfunc') {
+                               $available_hashes = Crypt::getAvailablePasswordHashes();
+                               if (!array_key_exists($result, $available_hashes)) {
+                                       $result = PASSWORD_DEFAULT;
+                                       self::Set('system.passwordcryptfunc', $result);
+                               }
+                       }
                }
+
                return $result;
        }
```

sro0:
`system.passwordcryptfunc` is set to `PASSWORD_BCRYPT`/`2y` when the error occurs.

The problem is that the PHP version used doesn't support Argon2. If Argon2 is unsupported the constants `PASSWORD_ARGON2I` and `PASSWORD_ARGON2ID` aren't defined in PHP -> exception is triggered e.g in at https://github.com/Froxlor/Froxlor/blob/dbf83c6f24b4dfdc6749eddc1b4c7e03878b5ae5/lib/Froxlor/Api/Commands/EmailAccounts.php#L160

d00p:
Yes, you're right. Totally missed that part - your fix seems exactly what is needed

PR #1233, opened 2024-01-29
Rename dovecot config file generated be renew hook to ensure it gets after default froxlor config file

sro0:
# Description

Currently generated file names by Froxlor for Dovecot using Debian Bookworm:
- /etc/dovecot/conf.d/99-froxlor.conf (from service config)
- /etc/dovecot/conf.d/99-froxlor-ssl.conf (from renew hook)

Dovecot includes all `conf.d/*.conf` files like this [dovecot.conf](https://github.com/dovecot/core/blob/35cba09cf91d79402f44ccdff8246c19caa2622e/doc/example-config/dovecot.conf#L92):
> The filenames are first sorted by their ASCII value and parsed in that order.

`.` == 0x2e
`-` == 0x2d

Thats why `99-froxlor-ssl.conf` gets parsed before `99-froxlor.conf` and the TLS values from renew hook might get overwritten.

# Solution
Choose filename for file generated by rename hook that gets parsed after default Froxlor config file.

# Type of change

**Test Configuration**:
* Distribution: Gentoo
* Dovecot: 2.3.20

# Checklist:


d00p:
Thanks for the contribution, although as stated in the code, it's not even final this way, as this might to be a setting (at least for the path). Just to be totally sure - would using `999-froxlor...` ensure it's read in last?

sro0:
Yes you are right, using `999-froxlor...` works. Just verified this.
Just depends if we want to break the 2 digit style that is used until now.

d00p:
We could also use `99-zzz-froxlor` but I think that's looks crappy :p

sro0:
Yeah whatever ^^. I just tripped over this while testing and this PR is a quick (and dirty) workaround.
As long as it is fixed in the final release (maybe with a setting for path), feel free to drop this PR.

d00p:
na i'll take it for now as it fixes the correct include order, thanks again :)

PR #1246, opened 2024-03-23
Fix "expires" option cannot have a year greater than 9999

halfgaar:
This fixes the exception: '"expires" option cannot have a year greater than 9999', which happens on upgrade from Debian 11 to 12. The session timeout in the DB is 9999999999999, so we constrain the value.

### DIFF
--- a/actions/admin/settings/110.accounts.php
+++ b/actions/admin/settings/110.accounts.php
@@ -35,6 +35,7 @@ return [
 					'varname' => 'sessiontimeout',
 					'type' => 'number',
 					'min' => 60,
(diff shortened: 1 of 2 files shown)


d00p (commented):
additonally, wouldn't it be wise to restrict entering values like `999999999999` to the field by adding a `max = '31536000',` to the config array in https://github.com/froxlor/Froxlor/blob/main/actions/admin/settings/110.accounts.php#L37

halfgaar:
Alright, amended the commit and pushed `--force-with-lease` over it.

PR #1251, opened 2024-05-07
Add config for Ubuntu 24.04 - Noble Numbat

rex2630:
Adding config for Ubuntu 24.04. Changed needed changes, like jammy to noble and 22.04 to 24.04. From my testing i didnt faced any other issues with the config of 22.04

### DIFF
--- a/lib/configfiles/noble
+++ b/lib/configfiles/noble
@@ -1,0 +1,4295 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<froxlor>
+	<distribution name="Ubuntu" codename="noble"
+		version="24.04" defaulteditor="/bin/nano">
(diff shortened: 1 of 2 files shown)


d00p:
I'll check these. Besides, the file needs to include the .xml extension

d00p (changes requested):
just two minor things (as far as I can see):

1) filename extensions needs to be .xml
2) default php on Noble is *8.3* - so a handful of 8.1 (former Jammy default) occurrences need to be adjusted to 8.3 (e.g. `a2dismod php8.1`)

d00p (approved):

rex2630:
Np and nice catch about the php version, i should check that, but as far im using nginx by default then adismod using hardcoded php version didn't even came to my mind

PR #1285, opened 2024-10-10
Add nginx HTTP/3 support

lukasbableck:
# Description
This PR adds support for HTTP/3 with nginx. 
It's not too different from HTTP/2, just a few more parameters and options in the vhost configs.
Support for HTTP/3 is only available in nginx 1.25+, which is currently only included in Ubuntu 24.10 (apparently this will be released just today :D), or of course can be installed via the packages provided by nginx directly.

\
There is only one small problem, which is why this PR is marked as a draft. 
Just like with HTTP/2, the `Host` header is not required and some clients do not include this header in their request.

This could be fixed by adding the following line to the `/etc/nginx/fastcgi_params` file:
```fastcgi_param HTTP_HOST $host;```
This works pretty well for me, but could maybe cause issues for others(?)

[`$host`](https://nginx.org/en/docs/http/ngx_http_core_module.html#var_host) uses the host name from the request line, or host name from the “Host” request header field, or the server name matching a request in this order.

Or, probably the preferred option, nginx "fakes" the Host header for HTTP/3 just like it does for HTTP/2 requests. See: https://mailman.nginx.org/pipermail/nginx-devel/2024-January/LCIUMLKCM2EBMEMTU3KXMW74AP2C4FYZ.html
However, this has not yet been implemented and I have also not found anything on whether this is actually planned or not.

## Type of change

# How Has This Been Tested?

Install nginx 1.25.0+, enable HTTP/3 in system settings, enable HTTP/3 per domain => clients should be able to use HTTP/3.

**Test Configuration**:

* Distribution: Debian 12
* Webserver: nginx 1.26.2 and 1.27.2 (current stable and mainline versions provided by nginx)
* PHP: 8.3 FPM 

# Checklist:

### CODE REVIEW

File: lib/formfields/admin/domains/formfield.domains_add.php [hunk 1]
```diff
@@ -234,6 +234,14 @@
 						'value' => '1',
 						'checked' => false
 					],
+					'http3' => [
+						'visible' => !empty($ssl_ipsandports) && Settings::Get('system.webserver') != 'lighttpd' && Settings::Get('system.http3_support') == '1',
```
d00p, line 238: this should be changed from
```diff
...
- Settings::Get('system.webserver') != 'lighttpd'
+ Settings::Get('system.webserver') == 'nginx'
...
```

as i doubt lighttp does this....

lib/formfields/admin/domains/formfield.domains_edit.php
```diff
@@ -266,6 +266,14 @@
 						'value' => '1',
 						'checked' => $result['http2']
 					],
+					'http3' => [
+						'visible' => !empty($ssl_ipsandports) && Settings::Get('system.webserver') != 'lighttpd' && Settings::Get('system.http3_support') == '1',
```
d00p, line 270: same as in formfield.domains_add.php

lib/formfields/customer/domains/formfield.domains_add.php
```diff
@@ -137,6 +137,14 @@
 						'value' => '1',
 						'checked' => false
 					],
+					'http3' => [
+						'visible' => $ssl_ipsandports && Settings::Get('system.webserver') != 'lighttpd' && Settings::Get('system.http3_support') == '1',
```
d00p, line 141: same as in admin/formfield.domains_add.php

lib/formfields/customer/domains/formfield.domains_edit.php
```diff
@@ -172,6 +172,14 @@
 						'value' => '1',
 						'checked' => $result['http2']
 					],
+					'http3' => [
+						'visible' => $ssl_ipsandports && Settings::Get('system.webserver') != 'lighttpd' && Settings::Get('system.http3_support') == '1',
```
d00p, line 176: same as in admin/formfield.domains_add.php

d00p (changes requested):
Nice work, just a minor adjustment, but very helpful, thank you

lib/formfields/admin/domains/formfield.domains_add.php
lukasbableck: Ah yeah, just copied this from http2 and Apache doesn't support h3. 🤦🏼

lukasbableck:
Since there has not been any changes to this in nginx yet I have been looking for another solution for the Host header problem.
Nginx recommends switching to the $host variable (there is actually a PR to change this in the default config https://github.com/nginx/nginx/pull/236), but to keep the same behavior as before for HTTP/1 and HTTP/2 something like the following could be used maybe. It will only use $http_host if it is not empty, otherwise $host will be passed to PHP.
LMK what you think @d00p 

```
fastcgi_param  HTTP_HOST 	  $host;
fastcgi_param  HTTP_HOST	  $http_host if_not_empty;
```

d00p (approved):

d00p:
> Please check the fix in this new pull request and patch :
> 
> [nginx/nginx#917](https://github.com/nginx/nginx/pull/917)

Don't quite get what it is we need to adjust now...?

lukasbableck:
Nothing. 
The linked PR would have implemented what I mentioned in the initial message in this PR:
> Or, probably the preferred option, nginx "fakes" the Host header for HTTP/3 just like it does for HTTP/2 requests

\
This apparently wont happen tho, so something like what I mentioned in https://github.com/froxlor/Froxlor/pull/1285#issuecomment-2629282016 still has to be configured to be able to use HTTP/3 with some applications. (Froxlor will always redirect to notice.html for example, as we already discussed briefly on Discord) 
I'd still prefer if Froxlor sets that configuration, but having to configure that manually if HTTP/3 should be used is fine for me personally too.

d00p:
Oh okay, when I read the https://github.com/nginx/nginx/pull/917 PR i thought it's been done and therefore fixed, too bad. I'll see what we can do

PR #1335, opened 2025-06-09
proposal to fix #1334

### LINKED ISSUE #1334: bind configuration format breaks on multi-line entries in DKIM key
After upgrading from 2.1 to 2.2.7, and force-recreating the bind zone files, bind fails to load any zone files containing DKIM entries with:
```
loading from master file /etc/bind/domains/***.zone failed: unbalanced quotes
```

Turns out, the public key in the database looks like:

````
-----BEGIN PUBLIC KEY-----
MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAtUvVsFAvprnMhS2SPfv4
mB/DodRMjMY+NOCu3N4ZLtwbp3GcNam+ShWYOdjkMBMkduAsNlHEo5+PAI/MiYbL
...
````
including newlines in the dkim_pubkey column values.

**Expected behavior**
The newlines should be removed from the string (or correctly quoted) before generating the TXT record from it.

I will provide a PR with a suggested fix.

### PR DESCRIPTION
**Pull request by skleber**
(bind configuration format breaks on multi-line entries in DKIM key)

Fixes #1334

## Type of change

Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/lib/Froxlor/Dns/Dns.php
+++ b/lib/Froxlor/Dns/Dns.php
@@ -549,7 +549,7 @@ class Dns
 			// start
 			$dkim_txt = 'v=DKIM1;';
 			// key
-			$dkim_txt .= 'k=rsa;p=' . trim($domain['dkim_pubkey']) . ';';
+			$dkim_txt .= 'k=rsa;p=' . trim(str_replace(PHP_EOL, '', $domain['dkim_pubkey'])) . ';';
 			// dkim-entry
 			$zone_dkim[] = $dkim_txt;
 		}

### CODE REVIEW

File: lib/Froxlor/Dns/Dns.php [hunk 1]
```diff
@@ -549,7 +549,7 @@ private static function generateDkimEntries(array $domain): array
 			// start
 			$dkim_txt = 'v=DKIM1;';
 			// key
-			$dkim_txt .= 'k=rsa;p=' . trim($domain['dkim_pubkey']) . ';';
+			$dkim_txt .= 'k=rsa;p=' . trim(str_replace(PHP_EOL, '', $domain['dkim_pubkey'])) . ';';
```
d00p, line 552: that would only fix half of the issue you are referring to, as it would still include `-----BEGIN PUBLIC KEY-----`and `-----END PUBLIC KEY-----` parts which should not even be there, as generated keys are generated via `rspamd` with parameter `-o plain` which only returns the key itself in one line. 

Still guess there was an update issue when you've upgraded from 2.1 to 2.2.x somehow

PR #1364, opened 2025-12-18
Remove `curl_close()` calls, has been not doing anything since 8.0 and is now officiall deprecated

bashgeek:
Fixes (as example):

```
PHP Deprecated: Function curl_close() is deprecated since 8.5, as it has no effect since PHP 8.0 in /var/www/froxlor/lib/Froxlor/Http/HttpClient.php on line 59
```


bashgeek:
It didn't work correctly before either, that's why it was pretty much abandoned back in 8.0. Even when it worked it did very little in terms of freeing resources - especially with a low-traffic and low-resource platform like Froxlor it's completely irrelevant, since nothing will get reused anyway. It's a non-breaking change, even before 8.0

PR #1387, opened 2026-02-25
Fix: "Allowed sender @domain” greift nicht, wenn die Absender-Mailbox existiert

MBungalski:
# Description

Adjusted templates for Debian and Ubuntu
Fix for Issue #1386

# Problem description
In Froxlor 2.3.4 you can allow domain-wide sending for a mailbox under “Allowed sender addresses” (e.g. @example.com), so that an SMTP login may send as any address within that domain.
In practice, this domain-wide permission does not apply as soon as the specific sender address exists as a real mailbox. In that case, the sender_login_maps lookup only returns the “owner” of the concrete address, and the domain-wide rule (@domain) is not additionally taken into account.

Result: sending is rejected even though @domain is configured as an allowed sender.

# Workaround / Patch (SQL change)
Updated the Debian config files so that the SQL query in /etc/postfix/mysql-virtual_sender_permissions.cf is extended with an additional lookup based on the domain part only:
`UNION (SELECT mail_sender_aliases.email FROM mail_sender_aliases WHERE mail_sender_aliases.allowed_sender = CONCAT('@', SUBSTRING_INDEX('%s','@',-1)))
`
## Type of change

# Checklist:

- [X] I have added tests that prove my fix is effective or that my feature works


MBungalski:
The scripts for `focal`, `jammy` and `noble` are now updated, too.

